PR #14, opened 2021-06-26
highlight.js

storopoli:
This is a PR to incorporate syntax highlighting to code blocks using [highlight.js](https://highlightjs.org/usage/).

I did already the extra script calls necessary in the `<head>` tag inside the `build()` function. I explicitely asked to download additional styles such as `julia` and `julia-repl`, and `python` and `python-repl`. We can have more than 180 different languagues, check [here](https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md) for aliases.

Now we need to add this to every code chunk that has a language.

```html
<pre><code class="language-alias">...</code></pre>
```

where alias would be the language alias. Probably the default should be `julia`. We would also want to add an option to flag a code output as `python` (since we are using it in Why Julia section).

I also included `julia-repl` and `python-repl` if we need syntax highlighting for those languages REPL.

I don't know how to do this in HTML, CSS. But I hoped that my initital efforts were of help.

If you could point me towards documentation or how to do it. I would glad do it. I think that this PR should also motivate a new feature add-on on Books.jl.


storopoli:
Ops, I only want to incorporate commit b0caa1d526f8a13971cd4190d7f66c5d9955662b. It pulled the whole train from the why-julia-2 branch...

storopoli:
closed in favor of #15

PR #15, opened 2021-06-26
Syntax Highlighting with highlight.js

storopoli:
This is a PR to incorporate syntax highlighting to code blocks using [highlight.js](https://highlightjs.org/usage/).

I did already the extra script calls necessary in the `<head>` tag inside the `build()` function. I explicitely asked to download additional styles such as `julia` and `julia-repl`, and `python` and `python-repl`. We can have more than 180 different languagues, check [here](https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md) for aliases.

Now we need to add this to every code chunk that has a language.

```html
<pre><code class="language-alias">...</code></pre>
```

where alias would be the language alias. Probably the default should be `julia`. We would also want to add an option to flag a code output as `python` (since we are using it in Why Julia section).

I also included `julia-repl` and `python-repl` if we need syntax highlighting for those languages REPL.

I don't know how to do this in HTML, CSS. But I hoped that my initital efforts were of help.

If you could point me towards documentation or how to do it. I would glad do it. I think that this PR should also motivate a new feature add-on on Books.jl.

### DIFF
--- a/src/JDS.jl
+++ b/src/JDS.jl
@@ -37,6 +37,16 @@ function build()
       function gtag(){dataLayer.push(arguments);}
       gtag('js', new Date());  gtag('config', 'G-2RKMTDS1S8');
     </script>
+
+    <!-- Highlight.js -->
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@11.0.1/build/styles/default.min.css">
+    <script src="https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@11.0.1/build/highlight.min.js"></script>
+    <!-- and it's easy to individually load additional languages -->
+    <script src="https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@11.0.1/build/languages/julia.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@11.0.1/build/languages/julia-repl.min.js"></script>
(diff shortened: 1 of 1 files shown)


rikhuijzer:
> Now we need to add this to every code chunk that has a language.

That would be possible. In Books.jl, it think it should go a long way if we change `code_block` from

    ```
    $code
    ```

to

    ```julia
    $code
    ```

However, we then still have no syntax highlighting in the PDF. Is that okay? Otherwise, we should look into Pandoc syntax highlighting: https://pandoc.org/MANUAL.html#syntax-highlighting, which will be more output independent.

Another thing is that it would be much better if we serve these Javascript files ourselves. It's very cumbersome to not be able to locally develop a project simply because internet is down. Also, for the end user it has two benefits:

1. The site is more quick, because there doesn't have to be a DNS lookup for the CDN. Note that it is not true that the files might be cached already since they have been used on another site. This caching is disabled by browsers for security reasons. Highlight.js will be downloaded for each site that uses it separately.
1. The user is not being tracked via the Javascript CDN.

storopoli:
We can serve them instead of downloading, totally agree.

We can start slow with syntax highlighting for the HTML version (which I think will be the most consumed). And then figure out how to incorporate syntax highlighting in PDF for Amazon.com (and Amazon.com.br) release(s).

rikhuijzer:
Okay sounds good. I‘ll pick this up soon and otherwise feel free to remind me

rikhuijzer:
Fixed by https://github.com/rikhuijzer/Books.jl/pull/164.

PR #18, opened 2021-06-27
Add code style blue

rikhuijzer:
So far, we're already complying with this style, but I think it would be good to mention it. It's a pretty nice style if you ask me.

Merge if you agree! Thanks


storopoli:
I agree. Great style to use!

PR #22, opened 2021-07-04
More Julia 101 Writing

storopoli:


storopoli:
@rikhuijzer I am having some issues with the Pandoc in the GHA CI. Could you take a look? My guess is that the error is in the PDF creation step, since the HTML has no issues whatsoever...

rikhuijzer:
Unfortunately, it's a font issue. From the CI log:

```
warning: could not represent character "ᵢ" (0x1d62) in font "[SourceSansPro-Regular.otf]/OT:script=latn;language=dflt;+pnum;mapping=tex-text;"
```

Is this something that you can work around? Otherwise, I'll make it my priority to set JuliaMono for the PDF generation.

storopoli:
I think we should set JuliaMono for the PDF generation. It would pair up with the HTML version and also would allow us to use unicode math fonts.

It is a big sellout that we propose in the Why Julia section. My opinion is to make the ᵢ available in the PDF.

rikhuijzer:
Agreed. My question was more about whether it should be fixed now or later. I'll fix it somewhere this week.

storopoli:
> Agreed. My question was more about whether it should be fixed now or later. I'll fix it somewhere this week.

I don't mind failed CIs for the PDF

contents/julia_101.md
```diff
@@ -423,55 +517,2108 @@ logarithm(10)
...
+
+Let's just show the first 4 lines of our downloaded file with the `readlines` function:
+
+```jl
+sco(
+"""
+readlines(my_file)[1:4]
+"""
+)
+```
```
storopoli, line 2621: @rikhuijzer Don't know to process this as a string... It looks really ugly.

contents/julia_101.md
```diff
@@ -423,55 +517,2108 @@ logarithm(10)
...
+Most of the time, specially when dealing with tabular data, we are using either one- or two-dimensional arrays.
+They are both `Array` types for Julia.
+But we can use the handy aliases `Vector` and `Matrix` for clear and concise syntax.
+
+#### Array Construction {#sec:array_construction}
+
+How do we construct an array?
+The simplest answer is to use the **default constructor**.
+It accepts the element type as the type parameter inside the `{}` brackets and inside the constructor you'll pass the element type followed by the desired dimensions.
+It is common to initialize vector and matrices with undefined elements by using the `undef` argument for type.
```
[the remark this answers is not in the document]
rikhuijzer: Basically, I would say that you can go perfectly well without using loops in Julia. This idea is also already sold by R, luckily

contents/julia_101.md
```diff
@@ -423,55 +517,2108 @@ logarithm(10)
...
+    """
+    )
+    ```
+
+* `hcat`: horizontal concatenation, a shorthand for `cat(...; dims=2)`
+
+    ```jl
+    sco(
+    """
+    hcat(ones(2), zeros(2))
```
rikhuijzer, line 1338: and maybe add a mention for `[a; b]`, because that is suggested [here](https://docs.julialang.org/en/v1/manual/style-guide/#Don't-overuse-...)

contents/julia_101.md
```diff
@@ -423,55 +517,2108 @@ logarithm(10)
...
+
+```jl
+sco(
+"""
+my_dict
+"""
+)
+```
+
+`Dict`s are also used in data manipulations by [`DataFrames.jl`](dataframes.html) and data visualization by [`Plots.jl`](plots.html).
```
rikhuijzer, line 1904: Maybe, also add the `Dict(zip(A, B))` where `A` and `B` are vectors, that is, iterable. It's a super useful constructor if you quickly want to have a mapping from `A` to `B`.

contents/julia_101.md
```diff
@@ -423,55 +517,2108 @@ logarithm(10)
...
+```
+
+## Filesystem {#sec:filesystem}
+
+In data science, most projects are undertaken in a collaborative effort.
+We share code, data, tables, figures and so on.
+Behind everything, there is the **operational system (OS) filesystem**.
+In an ideal work, code would run the *same* in *different* OS.
+But that is not what actually happens.
+Some OS (mostly Windows) have some quirks that can **break code** or **introduce undefined behavior**.
```
rikhuijzer, line 1938:
```suggestion
One instance of this is the difference of Windows paths, such as `C:\\user\john\`, and Linux paths, such as `/home/john`.
```

contents/julia_101.md
```diff
@@ -423,55 +517,2108 @@ logarithm(10)
...
+DateTime("1987-09-13T21:21:00", "yyyy-mm-ddTHH:MM:SS")
+"""
+)
+```
+
+You can find more on how to specify different format as strings in the [Julia `Dates`' documentation](https://docs.julialang.org/en/v1/stdlib/Dates/#Dates.DateFormat).
+Don't worry if you have to revisit it all the time, we ourselves have to do it all the time when working with dates and timestamps.
+
+According to [Julia `Dates`' documentation](https://docs.julialang.org/en/v1/stdlib/Dates/#Constructors), using the `Date(date_string,format_string)` function is fine if only called a few times.
+If there are many similarly formatted date strings to parse however, it is much more efficient to first create a `DateFormat` type, and then pass it instead of a raw format string.
```
rikhuijzer, line 2142: Would this be a real issue to our users? They can read about it in the documentation too

contents/julia_101.md
```diff
@@ -423,55 +517,2108 @@ logarithm(10)
...
+
+To begin we first import the `Random` module:
+
+```julia
+using Random
+```
+
+We have **two main functions that generate random numbers**:
+
+* `rand`: samples a **random element** of a data structure or type.
```
rikhuijzer, line 2401: Didn't you say that `rand` is in Julia Base? I can use it without using Random, and it doesn't even load more methods
```
julia> rand
rand (generic function with 69 methods)

julia> using Random

julia> rand
rand (generic function with 69 methods)
```

contents/julia_101.md
```diff
@@ -423,55 +517,2108 @@ logarithm(10)
...
+
+```jl
+sco(
+"""
+randn((2, 2))
+"""
+)
+```
+
+#### `seed!` {#sec:random_seed}
```
rikhuijzer, line 2524: yes, this is important

contents/julia_101.md
```diff
@@ -423,55 +517,2108 @@ logarithm(10)
...
+Random.seed!(123)
+rand(3)
+"""
+)
+```
+
+Note that `seed!` is not automatically exported by the `Random` module.
+We have to call it with the `Module.function` syntax.
+
+In order to avoid tedious and inefficient repetition of `seed!` all over the place we can instead define an instance of a `seed!` and pass it as a first argument of **either `rand` or `randn`**.
```
rikhuijzer, line 2555: Can't you just call `Random.seed!` once at the start of a program to make it all reproducible? As long as you don't change the logic of the program, nothing changes. For most science things, this is good enough, I would say? I'm not sure, tho

contents/julia_101.md
rikhuijzer: `sco("readlines(my_file)[1:4]"; process=string, post=output_block)`?

rikhuijzer (commented):
I'm very impressed by the amount of text that you have written! I hope you find my comments useful.

contents/julia_101.md
```diff
@@ -423,55 +517,2108 @@ logarithm(10)
...
+You begin with a keyword, in this case `for`.
+Then you specify what Julia should "loop" for, i.e. a sequence.
+Also, like everything else, you must finish with the `end` keyword.
+
+So, to make Julia print every number from 1 to 10, you'll need the following for loop:
+
+```jl
+sco(
+"""
+for i ∈ 1:10
```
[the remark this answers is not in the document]
storopoli: Yeah you are probably right. I, as a CS/Math guy, love the `∈`. But we should focus on our target audience. I will revert the `∈` to `in`.

contents/julia_101.md
```diff
@@ -423,55 +517,2108 @@ logarithm(10)
...
+For long strings we can use triple double quotes:
 
-```{=comment}
-the way to get data out of functions
+```jl
+sco(
+"""
+typeof(
+\"\"\"
+This is a big multiline string.
```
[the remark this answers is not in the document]
storopoli: Do we need to change the example? I don't understand.

contents/julia_101.md
```diff
@@ -423,55 +517,2108 @@ logarithm(10)
...
 
-### Named Tuple {#sec:namedtuple}
-### Array {#sec:array}
-### Dict {#sec:dict}
+#### String Concatenation {#sec:string_concatenation}
 
-## Filesystem {#sec:filesystem}
+One nice thing we can do with strings in Julia is **string concatenation**.
+Suppose you want to construct a new string that is the concatenation of two or more strings.
+This is accomplish in julia either with the `*` operator (this is where Julia departs from other scientific open-source programming languages) or the `join` function:
```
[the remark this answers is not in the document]
storopoli: It's right after this. I don't want to overload the reader with lots of alternatives.

contents/julia_101.md
storopoli: Too much complexity... Let's stick with hcat for now

contents/julia_101.md
storopoli: Ok done in 8867c5a8ff5e147361ba4b5d6dd36f85e4375fbe

contents/julia_101.md
storopoli: I decided to cover since we are using it in DataFrames as timeseries analysis.

contents/julia_101.md
storopoli: Done in d728d40

contents/julia_101.md
storopoli: Added in d728d40
Good call

contents/julia_101.md
```diff
@@ -423,55 +517,2108 @@ logarithm(10)
...
+"""
+logarithm.([1, 2, 3])
+"""
+)
+```
+
+#### Functions with a bang `!` {#sec:function_bang}
+
+In Julia syntax is is common to append `!` to names of functions that modify their arguments.
+This is a convention that warns the user that the function is **not pure**, i.e. it has *side effects*.
```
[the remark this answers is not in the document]
storopoli: added an example in a1b5b3d

contents/julia_101.md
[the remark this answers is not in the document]
storopoli: Done in 5e83871

contents/julia_101.md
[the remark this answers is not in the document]
storopoli: Good point. I've added something in d890d9f

contents/julia_101.md
```diff
@@ -423,55 +517,2108 @@ logarithm(10)
...
+s="Julia"
+
+my_quick_namedtuple = (; i, f, s)
+"""
+)
+```
+
+### Array {#sec:array}
+
+Arrays are a **systematic arrangement of similar objects, usually in _rows_ and _columns_**.
```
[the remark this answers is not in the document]
storopoli: I've added a phrase in 110f199

contents/julia_101.md
storopoli: Thanks

storopoli (commented):
Done!

PR #25, opened 2021-07-07
Fix broken PDF build on Julia 101 branch

rikhuijzer:
The CI should now pass thanks to https://github.com/rikhuijzer/Books.jl/pull/147.

### DIFF
--- a/contents/julia_101.md
+++ b/contents/julia_101.md
@@ -240,17 +240,17 @@ scob(
 )
 ```
 
-Regarding numeric comparison, Julia have three major types of comparisons:
+Regarding numeric comparison, Julia has three major types of comparisons:
 
 1. **Equality**: either something is *equal* or *not equal* another
-   * `==`: equality
-   * `!-` or `≠`: inequality
-2. **Less than**: either something is *less than* or *less than or equal to*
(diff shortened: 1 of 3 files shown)


rikhuijzer:
Okay, got it. The issue is that **inline code** with Unicode symbols ends up being handled by the listings package. For example, `σᵢ` is converted to `\passthrough{\lstinline!σᵢ!}` by Pandoc. Unfortunately, this is a very complex issue, see, for example, how the definitions look for Unicode listings at https://github.com/mossr/julia-mono-listings/blob/master/julia_listings_unicode.tex. I've tried to add the Julia listings from https://github.com/mossr/julia-mono-listings/, but this doesn't play nice with Tectonic. The reason that Tectonic is a must is because it handles dependencies extremely well and is provided via Yggdrasil.

So, in summary. I don't see any _reasonable_ solution from the Books.jl side. But, luckily, the solution is simply not to put Unicode inside inline code. This PR does that. This looks as follows:

# From the webpage:

image

# From the PDF:

image

# For the other situation, from the webpage:

image

storopoli (approved):
Yeah it is an elegant fix. LGTM.

PR #26, opened 2021-07-09
(Temporary) Remove hitcounter

rikhuijzer:
@storopoli The hitcounter takes 30 seconds to load and is, I think therefore, not visible

Mainly, I'm also making this change to verify that Books.jl is now fixed.


storopoli:
Agreeded! 😆

PR #28, opened 2021-07-11
Apply many small rewrites

rikhuijzer:
Many many small rewrites.


contents/julia_101.md
```diff
@@ -1976,35 +1993,50 @@ sco(
...
+```
+
+```jl
+sco("""
+sym = Symbol(s)
+""")
+```
+
+One simple benefit of symbols is that you have to type one character less, that is, `:some_text` versus `"some text"`.
+We use `Symbol`s a lot in data manipulitions with the `DataFrames.jl` package (@sec:dataframes) and data visualizations (@sec:plots and -@sec:makie).
```
storopoli, line 2038:
```suggestion
We use `Symbol`s a lot in data manipulations with the `DataFrames.jl` package (@sec:dataframes) and data visualizations (@sec:plots and -@sec:makie).
```

storopoli (approved):

rikhuijzer:
Thanks for the quick review!

Issue #31, opened 2021-07-20
Introduce `...` splat operator in Julia 101 Section


rikhuijzer:
In response to the comment in #29: I think that it is important enough to introduce in the intro.

storopoli:
It will go into Julia 101 Section

storopoli:
It also appears in #37

PR #37, opened 2021-07-25
Add section on `DataFrames.join`

rikhuijzer:


contents/indexing.md
```diff
@@ -94,8 +94,8 @@ sco("grades_2020()"; process=without_caption_label)
 Add ref to multiple dispatch in the intro
 ```
 
-We can filter rows by using `filter(f::Function, df)`.
-This function is very similar to the function `filter(f::Function, V::Vector)` from Julia itself.
+We can filter rows by using `filter(source => f::Function, df)`.
+Note how this function is very similar to the function `filter(f::Function, V::Vector)` from Julia itself.
```
lazarusA, line 98: why is this important? what's the different, is it worth saying?

contents/join.md
```diff
@@ -0,0 +1,140 @@
+## Join {#sec:join}
+
+At the start of this chapter, we showed multiple tables and gave questions related to multiple tables.
+However, we haven't talked about combining tables yet, which we will do in this section.
+In `DataFrames.jl`, combining multiple tables is done via _joins_.
+Joins are extremely powerful, but might take some time to wrap your head around.
+It is not necessary to know the joins below by heart, the [`DataFrames.jl` documentation](https://DataFrames.juliadata.org/stable/man/joins/) and this book list them too.
+But, it's good to know that they exists.
+If you ever find yourself looping over rows in a DataFrame and and comparing it with other data, then you probably need one of the joins below.
```
lazarusA, line 9: `in a DataFrame and and comparing`
double `and`. Remove one.

contents/join.md
```diff
@@ -0,0 +1,140 @@
...
+```jl
+s = "grades_2021()"
+sco(s; process=without_caption_label)
+```
+
+To do this, we are going to use joins.
+The `DataFrames.jl` lists no less than seven kinds of joins.
+We show them all here, because they are all useful.
+This first mentioned one is `innerjoin`.
+Suppose that we have two datasets `A` and `B` with respectively columns `A_1, A_2, ..., A_n` and `B_1, B_2, ..., B_m` **and** one of the columns has the same name, say `A_1` and `B_1` are both called `:id`.
```
lazarusA, line 29: I would say: Suppose that we have two datasets `A` and `B` with  columns `A_1, A_2, ..., A_n` and `B_1, B_2, ..., B_m` respectively **and** .... 

the `respectively` at the end, up to you.

contents/join.md
```diff
@@ -0,0 +1,140 @@
...
+For example, compare the output below to the previous output:
+
+```jl
+s = "leftjoin(grades_2021(), grades_2020(); on=:name)"
+sco(s; process=without_caption_label)
+```
+
+Lastly, we have the `semijoin` and `antijoin`.
+The semi join is even more restrictive than the inner join.
+It returns only the elements from the left DataFrame which are in both DataFrames:
```
lazarusA, line 111: so, this is the intersection from groups?

contents/select.md
```diff
@@ -0,0 +1,191 @@
+## Select {#sec:select}
+
+Whereas `filter` removes rows, `select` removes columns.
```
lazarusA, line 3: does is actually removing? or creating a new data array?, not a view, right?

storopoli: I think that `filter!` and `select!` might be worth to mention

contents/select.md
```diff
@@ -0,0 +1,191 @@
...
+
+For the age column, we have a similar problem:
+
+```jl
+s = "sort(wrong_types(), :age)"
+scsob(s)
+```
+
+This isn't right, because an infant is younger than adults and adolescents.
+The solution for this categorical data problem is to use `CategoricalArrays.jl`:
```
lazarusA, line 144: yet, another package. I feel sometimes that Julia is like a zoo, one needs to know what species is needed. This is just a general remark, people coming from other languages are not use to this. They kinda of expect to have everything loaded and working. Probably we could say something about this somewhere?

lazarusA:
Would it be nice to have an application of all these tools for analysing some real data? I do /used to have a data set from finance where all this was very useful.

contents/join.md
```diff
@@ -0,0 +1,140 @@
...
+## Variable Transformations
+
+```{=comment}
+manipulate variables
+DataFrames.transform
+Ifelse and case_when
+```
+
+In @sec:filter, we saw that `filter` works by taking one or more source columns and filtering it by applying a function.
+In other words, we saw that `source => f::Function` like in, for example, `filter(:name => name -> name == "Alice", df)`.
```
storopoli, line 135: DataFrames and bkamins call this "DataFrames minilanguage". Maybe we should introduce this term to the reader?

See here: https://bkamins.github.io/julialang/2020/12/24/minilanguage.html

contents/select.md
```diff
@@ -0,0 +1,191 @@
...
+
+Whereas `filter` removes rows, `select` removes columns.
+However, `select` is much more versatile than just removing columns, as we will discuss in this section.
+First, let's create a dataset with multiple columns:
+
+```jl
+@sco responses()
+```
+
+Here, the data represents answers on five questions (`q1`, `q2`, ..., `q5`) in a questionnaire.
```
storopoli, line 11:
```suggestion
Here, the data represents answers for five questions (`q1`, `q2`, ..., `q5`) in a questionnaire.
```

I think is "answers for" not "answers on". But my grammar even in Portuguese sucks...

contents/select.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+We can also use strings if we want:
+
+```jl
+s = """select(responses(), "id", "q1", "q2")"""
+sco(s, process=without_caption_label)
+```
+
+To select everything _except_ one or more columns, use `Not` with either a single column name:
```
storopoli, line 27: The `Not` `Between` etc are called column selectors in DataFrames.jl. If you type `?select` in julia with the `using DataFrames` you get:

```julia
standard column selectors (integers, Symbols, strings, vectors of integers, vectors of Symbols, vectors of
       strings, All, Cols, :, Between, Not and regular expressions)
```

Maybe we should introduce this term to the reader?

contents/select.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+Even renaming columns is possible via `select` via `source => target`:
+
+```jl
+s = """select(responses(), 1 => "participant", :q1 => "age", :q2 => "nationality")"""
+sco(s, process=without_caption_label)
+```
+
+which, thanks to the `...` operator, we can also write as:
```
storopoli, line 74: It is already the second time we are mentioning the splat operator `...`. I think I need to include it in the Julia chapter. What do you guys think? @lazarusA and @rikhuijzer

contents/select.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+Because the date column has the wrong type, sorting won't work correctly:
+
+```jl
+s = "sort(wrong_types(), :date)"
+scsob(s)
+```
+
+To fix the sorting, we can use `Date` as described in @sec:dates:
```
storopoli, line 120:
```suggestion
To fix the sorting, we can use `Date` from Julia's standard library as described in @sec:dates:
```

contents/select.md
storopoli, line 144: I think Categorical data and `CategoricalArrays.jl` should have a section for itself

contents/select.md
```diff
@@ -0,0 +1,191 @@
...
+s = """
+    df = correct_types()
+    a = df[1, :age]
+    b = df[2, :age]
+    a < b
+    """
+scob(s)
+```
+
+which would have been wrong for strings:
```
storopoli, line 185:
```suggestion
which would give wrong comparisons when the element type were strings:
```

storopoli (changes requested):
Some suggestions and discussions.

contents/select.md
lazarusA: Yes, I also use it a lot for some plots. 👍🏼

storopoli: Ok. gonna put in the TODO list with an issue and assign to me

contents/join.md
rikhuijzer: Yeah, I did that later in the text. Here, the full `source => transformation => target` wasn't used. I'll leave this one for you to figure out on your pass over this chapter :yum:

contents/select.md
rikhuijzer: > not a view, right?

Yes, you're correct. I'll try to word it a bit more careful.

> I think that filter! and select! might be worth to mention

I'll do it in the next round :+1:

contents/select.md
rikhuijzer: https://github.com/JuliaDataScience/JuliaDataScience/issues/31

contents/select.md
rikhuijzer: Compared to R or Python, it's not so much worse, I think. There, they even create a package which groups all the packages :rofl: (tidyverse). 

Compared to C++ or C, it indeed is worse, but in C++ you would just implement everything yourself because managing dependencies is a huge struggle.

> Probably we could say something about this somewhere?

You can open an issue if you want :smile:

contents/indexing.md
rikhuijzer: Because it explains why `filter(..., df)` and `select(df, ...)`, `subset(df, ...)` and `transform(df, ...)`.

contents/select.md
rikhuijzer: #54

contents/select.md
rikhuijzer: #54

rikhuijzer:
> Would it be nice to have an application of all these tools for analysing some real data? I do /used to have a data set from finance where all this was very useful.

Have you looked at the output? I've chosen the data specifically to show how the methods work. With real-world data, the examples would be, I think, much more difficult to understand. It's a

lazarusA:
> Have you looked at the output? I've chosen the data specifically to show how the methods work. With real-world data, the examples would be, I think, much more difficult to understand.

Yes. What I mean its like a one script where all _join_ methods are shown in action. It's just a suggestion. Probably I could show you one later.

Issue #23, opened 2021-07-05
Notation discussion points


rikhuijzer:
Notation discussion points from #20:

1. Always using : before the start of a code block.
1. Mentioning functions like `DataFrame` as `DataFrame()` or `DataFrame(...)`.
1. My suggestion: only Julia objects between backticks and filenames and extension names between quotation marks (like Julia's strings).

storopoli:
1. agreed!
2. I prefer `DataFrame()` without the `...` which would imply something being splatted which in turn can generate confusion and ambiguity
3. I did not understood this, could you give an example?

rikhuijzer:
2. Okay, fair that `...` is not a good solution. Yet, I think that `DataFrame()` or `CSV.write()` causes similar confusion. In code, it would mean calling the function without arguments. For example,
```
julia> using CSV

julia> CSV.write()
ERROR: MethodError: no method matching write()
Closest candidates are:
  write(::Tables.Schema{names, types} where types, ::Any, ::Any, ::Any; append, header, bufsize) where names at /home/rik/.julia/packages/CSV/Zl2ww/src/write.jl:195
  write(::Nothing, ::Any, ::Any, ::Any; append, header, bufsize) at /home/rik/.julia/packages/CSV/Zl2ww/src/write.jl:223
  write(::Any; kwargs...) at /home/rik/.julia/packages/CSV/Zl2ww/src/write.jl:134
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[10]:1
```
which can be very confusing to beginners, I think. In the Julia documentation, references to methods are just via backticks, so it would be `DataFrame` or `CSV.write`. For instance, 

"If you want to check that each converted value is a valid code point, use the `isvalid` function:" ([source](https://docs.julialang.org/en/v1/manual/strings/)) and

" It also provides a single general outer `Point` constructor that takes pairs of real arguments" ([source](https://docs.julialang.org/en/v1/manual/constructors/)).

3. You suggested in the PR to change something about the CSV extension and filename from "myfile.csv" to `myfile.csv`. I would prefer the former, due to it being more consistent with the Julia code since "myfile.csv" is a String and not some Julia object like a struct, function or method. We could also do a combination of your and my suggestion, namely `"myfile.csv"`.

storopoli:
2. Let's follow Julia syntax. So all functions would be referenced as `function` then
3. Let's follow also Julia syntax. So "myfile" instead of the `"myfile`"

[...]

PR #44, opened 2021-08-02
Add section on notation

rikhuijzer:
Moves our discussion from #23 into the book/into git.

### DIFF
--- a/contents/notation.md
+++ b/contents/notation.md
@@ -1,0 +1,68 @@
+## Notation {#sec:notation}
+
+In this book, we try to keep notation as consistent as possible.
+This makes reading and writing code easier.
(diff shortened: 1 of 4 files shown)


contents/notation.md
```diff
@@ -57,7 +57,7 @@ From the style guide, we attempt to adhere specifically to:
   ```
 - Do not omit zeros in floats (even though Julia allows it).
   Hence, write `1.0` instead of `1.` and write `0.1` instead of `.1`.
-- Use `in` in for loops and not `=` or `∈` (even though Julia allows it).
```
storopoli: Ok!

storopoli (approved):

PR #45, opened 2021-08-05
Review dataVis

rikhuijzer:


rikhuijzer:
Oops. Review is not necessary now. This PR is still a work in progress.

lazarusA (approved):
Thanks for your input. Looks good to me. Now it reads more clearly. Thanks. This rule about breaking the line when the sentence finishes I'm still not sure about it. I was trying to have lines not longer than 92 characters per line. Would that not be better?

rikhuijzer:
> Thanks for your input. Looks good to me. Now it reads more clearly. Thanks. This rule about breaking the line when the sentence finishes I'm still not sure about it. I was trying to have lines not longer than 92 characters per line. Would that not be better?

Good point. We're inconsistent currently. Let's discuss on Friday.

PR #58, opened 2021-08-09
added draft for color and colormaps in Makie

lazarusA:
please, feedback.


src/makie.jl
```diff
@@ -162,4 +162,65 @@ function multiple_example_themes()
...
+    # Epicycloid lines 
+    x(r,k,θ) = r*(k .+ 1.0).*cos.(θ) .- r*cos.((k .+ 1.0) .* θ)
+    y(r,k,θ) = r*(k .+ 1.0).*sin.(θ) .- r*sin.((k .+ 1.0) .* θ)
+    θ = LinRange(0,6.2π,1000)
+    
+    axis = (; xlabel = L"x(\theta)", ylabel = L"y(\theta)", 
+        title = "Epicycloid", aspect= DataAspect())
+    figure = (;resolution = (500,400), font= "CMU Serif")
+
+    fig, ax, _ = lines(x(1,1,θ), y(1,1,θ); color = "firebrick1", #string
```
rikhuijzer, line 177:
```suggestion
    fig, ax, _ = lines(x(1,1,θ), y(1,1,θ); color = "firebrick1", # string 
```

rikhuijzer (commented):
These plots look insane again! I love the `scatters_and_lines()` figure.

For a later draft, maybe you could also explain to the reader how these figures are created. Basically, you do some trial and error, right? 

Also, the code here looks very daunting, which is a bit inherent to plotting, I think. It would be nice if we can find ways to simplify things.

Lastly, I think the size of the Makie.jl chapter is getting quite big. If it's up to me, it's worth making it a chapter instead of a section.

storopoli:
I don't know how to comment since it was from a different merged PR but in L128 of `data_vis.md`

```julia
or as a `Dict` calling `pltobject.attributes.attributes`.
```

isn't it duplicated? It is not only `pltobject.attributes`?

lazarusA:
@storopoli with `pltobject.attributes` you get the list shown in the book, which I kinda like more than the other that is mentioned as a `Dict`. So both, are correct.

lazarusA:
@rikhuijzer Yes, explaining the creation of each plot where necessary will also be taking care of, once the whole chapter if complete. Let's call it filling the gaps :D 
- Yes is a little bit long, specially the last ones. This was my second iteration, the first version was even worst :D, I will think about it.
- Yes, probably it will be better to have an independent chapter, things are getting longer. :D

contents/data_vis.md
```diff
@@ -324,7 +324,7 @@ sco(s)
 
 Now, let's move on and do a plot with LaTeX strings and a custom theme.
 
-### using LaTeXStrings
+### using LaTeXStrings.jl
 
 LaTeX support in Makie is also available.
```
storopoli, line 329:
```suggestion
LaTeX support in `Makie.jl` is also available.
```

I think that we need to be consistent with previous subsections.

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
 Well, what about different default colors or palettes.
 In our next section, we will see how to use again [Cycles](http://makie.juliaplots.org/stable/theming.html#Cycles) and know a little bit more about them, plus some additional keywords in order to achieve this.
```
storopoli, line 369:
```suggestion
In our next section, we will see how to use again [`Cycles`](http://makie.juliaplots.org/stable/theming.html#Cycles) and know a little bit more about them, plus some additional keywords in order to achieve this.
```

more consistency since it is a type

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
...
+Using [Colors.jl](https://github.com/JuliaGraphics/Colors.jl) is supported in `Makie.jl`
+so that you can used [names colors](https://juliagraphics.github.io/Colors.jl/latest/namedcolors/) or pass `RGB` or `RGBA` values.
+Regarding colormaps, all those that work with [Plots.jl](https://github.com/JuliaPlots/Plots.jl) also do here.
+Additionally, colormaps from [ColorSchemes.jl](https://github.com/JuliaGraphics/ColorSchemes.jl) and [PerceptualColourMaps.jl](https://github.com/peterkovesi/PerceptualColourMaps.jl) can also be used.
+Worth knowing now, your can reverse a colormap by doing `Reverse(:colormap_name)`.
+And obtain a transparent color or colormap with `color= (:red,0.5)` and `colormap = (:viridis, 0.5)`.
+
+Different use cases will be shown next. Then we will difine a custom theme with new colors and a colorbar palette.
+
+By default `Makie.jl` has a predifined set of colors in order to cycle trought them automatically.
```
storopoli, line 383:
```suggestion
By default `Makie.jl` has a predefined set of colors in order to cycle trough them automatically.
```

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
...
+As shown in the previuos figures, where no specific color was set.
```
storopoli, line 384:
```suggestion
As shown in the previous figures, where no specific color was set.
```

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
...
+sco(s2color2)
+```
+
+
+ **Custom cycle**
+
+Here, we could define a global `Theme` with a new cycle for colors, however that is not the recommend way to do it.
+It's better to define a new theme and use as shown before.
+Lets define a new one with a `cycle` for `:color`, `:linestyle`, `:marker` and a new colormap default.
+Lets add this new attributes to our previuos `publication_theme`.
```
storopoli, line 498:
```suggestion
Lets add this new attributes to our previous `publication_theme`.
```

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
...
+    fig
+    label = "Reverse_colormap_sequential" # hide
+    caption = "Reverse colormap sequential and colorrange." # hide
+    Options(fig; filename=label, label, caption) # hide
+    """
+sco(scolor)
+```
+
+When setting a `colorrange` usually the values outside this range are colored with the first and last color from  the colormap.
+However, sometimes is better to specified the color that you want at both ends. We do that with `highclip` and `lowclip`:
```
storopoli, line 418:
```suggestion
However, sometimes is better to specify the color you want at both ends. We do that with `highclip` and `lowclip`:
```

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
...
+```jl
+@sco JDS.set_colors_and_cycle()
+```
+
+Where, in the first two lines we have used the keyword `color` to specified our color.
+The rest its using the default cycle set of colors. Later, we will learn how to do a custom cycle.
+
+Regarding colormaps, we are already familiar with the keyword `colormap` for heatmaps and scatters.
+Here, we show that a colormap can also be specified via a `Symbol` or a `String`, similar to colors.
+Or, even a vector of RGB colors.
```
storopoli, line 397:
```suggestion
Or, even a vector of `RGB` colors.
```

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
...
+    Colorbar(fig[1,2], pltobj, label = "Diverging colormap")
+    fig
+    label = "diverging_colormap" # hide
+    caption = "Diverging Colormap with low and high clip." # hide
+    Options(fig; filename=label, label, caption) # hide
+    """
+sco(s)
+```
+
+But we mentioned that also RGB vectors are valid options.
```
storopoli, line 441:
```suggestion
But we mentioned that also `RGB` vectors are valid options.
```

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
...
+For our next example you could pass the custom colormap _perse_ or use `cgrad` to force a categorical colorbar.
```
storopoli, line 442:
```suggestion
For our next example you could pass the custom colormap _perse_ or use `cgrad` to force a categorical `Colorbar`.
```

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
...
+    label = "categorical_colormap" # hide
+    caption = "Categorical Colormap." # hide
+    Options(fig; filename=label, label, caption) # hide
+    """
+sco(scat)
+```
+
+Lastly, the ticks in the colorbar for the categorial case are not centered by default in each color.
+This is fixed by passing custom ticks, as in `cbar.ticks = (positions, ticks)`.
+The last situation is when passing a `tuple` of two colors to `colormap` as symbols, strings or a mix.
```
storopoli, line 470:
```suggestion
The last situation is when passing a tuple of two colors to `colormap` as symbols, strings or a mix.
```

vector above do not have a backtick, so this probably shouldn't also...

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
...
+You will get an interpolated colormap between these two colors.
+
+Also, `hex` coded colors are also accepted. So, on top or our heatmap let's put one semi-transparent point using this.
```
storopoli, line 473: Do we need a backtick in `hex`? Is is a type or function from Julia/Makie etc?

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
...
+    fig
+    label = "colormap_two_colors" # hide
+    caption = "Colormap from two colors." # hide
+    Options(fig; filename=label, label, caption) # hide
+    """
+sco(s2color2)
+```
+
+
+ **Custom cycle**
```
storopoli, line 493: Isn't this better to become a sub-header for the section instead of bold?

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
...
+
+Here, we could define a global `Theme` with a new cycle for colors, however that is not the recommend way to do it.
```
storopoli, line 495:
```suggestion
Here, we could define a global `Theme` with a new cycle for colors, however that is **not the recommend way** to do it.
```

Always have to be careful with these not stuff in the middle of the sentence. I have high-school traumas from nasty questions.

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
...
+It's better to define a new theme and use as shown before.
+Lets define a new one with a `cycle` for `:color`, `:linestyle`, `:marker` and a new colormap default.
```
storopoli, line 497:
```suggestion
Lets define a new one with a `cycle` for `:color`, `:linestyle`, `:marker` and a new `colormap` default.
```

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
...
+Lets add this new attributes to our previuos `publication_theme`.
+
+```jl
+@sc new_cycle_theme()
+```
+
+And apply it to a plotting function to see it in action.
```
storopoli, line 504:
```suggestion
And apply it to a plotting function to see it in action.
```

It is the third time you are using "see it in action". I think we should change to another phrasing with the same meaning. What you think?

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
...
+s = """
+    with_theme(scatters_and_lines, new_cycle_theme())
+    label = "custom_cycle" # hide
+    caption = "Custom theme with new cycle and colormap." # hide
+    Options(current_figure(); filename=label, caption, label) # hide
+    """
+sco(s)
+```
+
+At this point you should be able to have complete control over your colors, line styles, markers and colormaps for your plots.
```
storopoli, line 520:
```suggestion
At this point you should be able to have **complete control** over your colors, line styles, markers and colormaps for your plots.
```

contents/data_vis.md
```diff
@@ -368,9 +368,157 @@ What more can we ask for?
...
+Next, we will like to manage and control layouts at will as well.
```
storopoli, line 521:
```suggestion
Next, we will dive into how to manage and control **layouts**.
```

storopoli (changes requested):
Some suggestions

contents/data_vis.md
lazarusA: it's a function from colours.

contents/data_vis.md
lazarusA: Yes, I tried that, it dosen't look good at this level. Probably it we make Makie a chapter, then it will.

contents/data_vis.md
lazarusA: Yes. Suggestions?

storopoli: Let's dive in.

Here is a good example.


etc

lazarusA:
All notes and suggestion were incorporated.

Issue #54, opened 2021-08-08
Issues in the DataFrames section


rikhuijzer:
- [ ] Filter doesn't "remove rows".
- [ ] Introduce column selectors and `?select`.
- [ ] CategoricalArrays.jl should have a section for itself.
- [ ]

storopoli:
I can take over this issue. But what is:
> Filter doesn't "remove rows".

PR #59, opened 2021-08-13
Add transform section

rikhuijzer:


contents/transform.md
```diff
@@ -0,0 +1,76 @@
...
+In this section, we discuss how to **transform** variables, that is, how to modify data.
+This goes via `source => transformation => target`.
+
+As a generous example, given the `grades_2020` dataset:
+
+```jl
+@sco process=without_caption_label grades_2020()
+```
+
+we can increase all the grades in `grades_2020`:
```
lazarusA, line 21:
```suggestion
we can increase all the grades in `grades_2020` by 1:
```

contents/transform.md
```diff
@@ -0,0 +1,76 @@
...
+Here, the `plus_one` function receives the whole `:grade_2020` row and not just one element as one might expect.
+That is the reason why we've added the dot before the plus symbol.
+The dot, in Julia's syntax, means broadcasting and allows Julia to add one to all the elements in `grades`.
+For instance:
+
+```jl
+sco("[1, 2, 3] .+ 1")
+```
+
+Like said above, the `DataFrames.jl` minilanguage talks is `source => transformation => target`.
```
lazarusA, line 40:
```suggestion
Like said above, the `DataFrames.jl` minilanguage talk is `source => transformation => target`.
```

contents/transform.md
```diff
@@ -0,0 +1,76 @@
...
+s = """
+    plus_one(grades) = grades .+ 1
+    transform(grades_2020(), :grade_2020 => plus_one)
+    """
+sco(s; process=without_caption_label)
+```
+
+Here, the `plus_one` function receives the whole `:grade_2020` row and not just one element as one might expect.
+That is the reason why we've added the dot before the plus symbol.
+The dot, in Julia's syntax, means broadcasting and allows Julia to add one to all the elements in `grades`.
```
lazarusA, line 33:
```suggestion
The `dot`, in Julia's syntax, means broadcasting and allows Julia to add one to all the elements in `grades`.
```

rikhuijzer:
I'm running out of ideas with this chapter. @storopoli, do you want to take over DataFrames from here?

storopoli:
Sure I can do that major review/edit that we've spoke about.

contents/transform.md
```diff
@@ -0,0 +1,94 @@
...
+
+```jl
+s = """
+    plus_one(grades) = grades .+ 1
+    transform(grades_2020(), :grade_2020 => plus_one)
+    """
+sco(s; process=without_caption_label)
+```
+
+Here, the `plus_one` function receives the whole `:grade_2020` row and not just one element as one might expect.
```
storopoli, line 29:
```suggestion
Here, the `plus_one` function receives the whole `:grade_2020` column and not just one element as one might expect.
```

I think is column right?

contents/transform.md
```diff
@@ -0,0 +1,94 @@
...
+That is the reason why we've added the dot before the plus symbol.
```
storopoli, line 30:
```suggestion
That is the reason why we've added the broadcasting "dot" `.` before the plus symbol.
```

Back in Julia 101 I've used the following phrasing:
> familiar broadcasting "dot" operator `.`

contents/transform.md
```diff
@@ -0,0 +1,94 @@
...
+The `dot`, in Julia's syntax, means broadcasting and allows Julia to add one to all the elements in `grades`.
```
storopoli, line 31:
```suggestion
The `.`, in Julia's syntax, means broadcasting and allows Julia to add one to all the elements in `grades`.
```

contents/transform.md
```diff
@@ -0,0 +1,94 @@
...
+For instance:
+
+```jl
+sco("[1, 2, 3] .+ 1")
+```
+
+Like said above, the `DataFrames.jl` minilanguage is `source => transformation => target`.
+So, if we want to have only one column in the output, we can do:
```
storopoli, line 39:
```suggestion
So, if we want to keep the naming of the `target` column in the output, we can do:
```


I think it's easier to say that if "we want to keep the naming of the target column".

Because the example before without the `=> target` works fine also.

Also you'll probably need to cover

```julia
transform(df, ...; renamecols=false)
```

which I use a lot.

contents/transform.md
```diff
@@ -0,0 +1,94 @@
...
+
+```jl
+s = """
+    transform(grades_2020(), :grade_2020 => plus_one => :grade_2020)
+    """
+sco(s; process=without_caption_label)
+```
+
+The same thing can also be written with `select` as follows:
```
storopoli, line 48:
```suggestion
The same transformation can also be written with `select` as follows:
```

or operation, or anything else besides "thing" 😄

contents/transform.md
```diff
@@ -0,0 +1,94 @@
...
+
+```jl
+s = """
+    select(grades_2020(), :, :grade_2020 => plus_one => :grade_2020)
+    """
+sco(s; process=without_caption_label)
+```
+
+where the `:` means "select all the columns" as described in @sec:select.
+Alternatively, it can be written by using Julia's broadcasting and modify column `grade_2020` by using `df.grade_2020`:
```
storopoli, line 58:
```suggestion
Alternatively, it can be written by using Julia's broadcasting and modify column `grade_2020` by acessing the column with `df.grade_2020`:
```

contents/transform.md
```diff
@@ -0,0 +1,94 @@
...
+```jl
+s = """
+    df = grades_2020()
+    df.grade_2020 = plus_one.(df.grade_2020)
+    df
+    """
+sco(s; process=without_caption_label)
+```
+
+But, although the last example is easier since it builds on more native Julia operations, we would advise to use the functions provided by `DataFrames.jl` in most cases.
```
storopoli, line 69: Why we would advise that?

I know why but our reader doesn't know and this is kinda of hand-waving argument...

contents/transform.md
```diff
@@ -0,0 +1,94 @@
...
+
+To show how to process two columns at the same time we take the left joined data from @sec:join:
```
storopoli, line 71:
```suggestion
To show how to transform two columns at the same time, we use the left joined data from @sec:join:
```

I think that is better to use transform than process since this is the name of the section.

contents/transform.md
```diff
@@ -0,0 +1,94 @@
...
+
+```jl
+s = """
+    leftjoined = leftjoin(grades_2020(), grades_2021(); on=:name)
+    """
+sco(s; process=without_caption_label)
+```
+
+With this, we can add a column saying whether someone passed:
```
storopoli, line 80:
```suggestion
With this, we can add a column saying whether someone was approved:
```

passed is too ambiguous

storopoli (changes requested):
Some comments

rikhuijzer:
@storopoli thanks for the nice suggestions. Was very easy to apply them again.

I've added one of your suggestions to #54. Would you like to pickup this issue? Then I can spend some time on fixing some (long standing) issues in Books.jl such as the logo in the top left.

PR #64, opened 2021-08-15
Added splatting operator to Julia 101

### LINKED ISSUE #31: Introduce `...` splat operator in Julia 101 Section

storopoli:
Since the splatting operator is mostly used in "collections" (such as arrays, vectors and tuples), I've included it in the bottom of the "Native Data Structures" section of "Julia 101" chapter.

Closes #31.

### DIFF
--- a/contents/julia_101.md
+++ b/contents/julia_101.md
@@ -2037,6 +2037,44 @@ sym = Symbol(s)
 One simple benefit of symbols is that you have to type one character less, that is, `:some_text` versus `"some text"`.
 We use `Symbol`s a lot in data manipulations with the `DataFrames.jl` package (@sec:dataframes) and data visualizations (@sec:plots and -@sec:makie).
 
+### Splat Operator {#sec:splat}
+
+In Julia we have the "splat" operator `...` which is mainly used in function calls as a **sequence of arguments**.
+We will occasionally use splatting in some function calls in **data manipulation** and **data visualization** chapters.
+
+The `...` operator takes a collection (often an array, vector, tuple or range) and converts into a sequence of arguments.
+The `...` is included after the collection that we want to "splat" into a sequence of arguments.
+For example, if we call a function on a range followed by `...`, it is the same as calling a function on all arguments of a collection.
+Syntactically speaking the following are the same:
+
+```julia
+collection = [x, y, z]
+
+function(collection...) = function(x, y, z)
+```
+
+Anytime Julia sees a splatting operator inside a function call, it will be converted on a sequence of arguments for all elements of the collection separated by commas.
+To understand the `...` operator, let's make a small example with a function that adds all elements in a collection:
+
+```jl
+scob("""
+function add_elements(xs...)
+    return reduce(+, xs)
+end
+
+add_elements([1, 2, 3]...)
+""")
+```
+
+It also works for ranges:
+
+```jl
+scob("""
+add_elements(1:2:20...)
+""")
+```
+
+
 ## Filesystem {#sec:filesystem}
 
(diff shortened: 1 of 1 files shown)


contents/julia_101.md
```diff
@@ -2037,6 +2037,43 @@ sym = Symbol(s)
...
 We use `Symbol`s a lot in data manipulations with the `DataFrames.jl` package (@sec:dataframes) and data visualizations (@sec:plots and -@sec:makie).
 
+### Splat Operator {#sec:splat}
+
+In Julia we have the "splat" operator `...` which is mainly used in function calls as a **sequence of arguments**.
+We will occasionally use splatting in some function calls in **data manipulation** and **data visualization** chapters.
+
+The `...` operator takes a collection (often an array, vector, tuple or range) and converts into a sequence of arguments.
+The `...` is included after the collection that we want to "splat" into a sequence of arguments.
+For example, if we call a function on a range followed by `...`, it is the same as calling a function on all arguments of a collection.
```
rikhuijzer, line 2047: 3blue1brown advises to give an example first and then discuss the nitty gritty details. I think this would be suitable here, because I'm afraid that I don't understand this sentence 😬

contents/julia_101.md
```diff
@@ -2037,6 +2037,43 @@ sym = Symbol(s)
...
+function(collection...) = function(x, y, z)
+```
+
+Anytime Julia sees a splatting operator inside a function call, it will be converted on a sequence of arguments for all elements of the collection separated by commas.
+To understand the `...` operator, let's make a small example with a function that adds all elements in a collection:
+
+```jl
+scob("""
+function add_elements(xs...)
+    return reduce(+, xs)
```
rikhuijzer, line 2062: Reduce is a very difficult function to understand for non-programmers, I think.

Can we write a, b, c out explicitly?

rikhuijzer (approved):
Slurping should also be added now or later (https://docs.julialang.org/en/v1/manual/faq/#The-two-uses-of-the-...-operator:-slurping-and-splatting)

contents/julia_101.md
storopoli: > Slurping should also be added now or later ([docs.julialang.org/en/v1/manual/faq/#The-two-uses-of-the-...-operator:-slurping-and-splatting](https://docs.julialang.org/en/v1/manual/faq/#The-two-uses-of-the-...-operator:-slurping-and-splatting))

Do we really need to cover slurping? the two cases that `...` appeared were splatting inside function calls in both dataframes and datavis chapters...

Will our user need to know the two cases slurping/splatting?

I am asking this because I want to justify every new increase of complexity.

contents/julia_101.md
storopoli: I can do this, see if it is better:

```julia
function add_elements(xs...)
    return +(xs...)
end
```

contents/julia_101.md
storopoli: > 3blue1brown advises to give an example first and then discuss the nitty gritty details. I think this would be suitable here, because I'm afraid that I don't understand this sentence

I will try to do the example first then and see if it makes it clearer.

rikhuijzer: Oh yes, you are very right. Slurping is not as important!

contents/julia_101.md
rikhuijzer: haha but now you have slurping

My suggestion was to make it explicit `add_elements(a, b, c) = a + b + c`. Everyone knows plus as being an infix operator.

rikhuijzer:
LGTM!

PR #83, opened 2021-08-30
Fix broken link to makie docs

### LINKED ISSUE #82: Chapter 7 Link broken to Makie Docs
There is a link broken in Chapter 7 `datavisMakie.md`:

In the "See Makie’s documentation for more." It redirects to http://makie.juliaplots.org/stable/backends_and_output.html#Backends-and-Output which is broken.

cc @lazarusA

lazarusA:
this fixes issue https://github.com/JuliaDataScience/JuliaDataScience/issues/82

Closes #82.

### DIFF
--- a/contents/data_vis_makie.md
+++ b/contents/data_vis_makie.md
@@ -13,7 +13,7 @@ It is reexported by every backend _(the machinery behind-the-scenes making the f
 There are three main backends which concretely implement all abstract rendering capabilities defined in Makie.
 One for non-interactive 2D publication-quality vector graphics: `CairoMakie.jl`.
 Another for interactive 2D and 3D plotting in standalone `GLFW.jl` windows (also GPU-powered), `GLMakie.jl`.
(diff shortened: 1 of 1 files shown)


rikhuijzer (approved):

rikhuijzer:
I've changed the text so that the issue got linked. For details, see https://huijzer.xyz/posts/github-gitlab-commands-cheatsheet/.

storopoli:
Aren’t we better linking to a static thing like dev or stable instead of a specific version?

rikhuijzer:
> Aren’t we better linking to a static thing like dev or stable instead of a specific version?

It is quite stable now, isn't it? Dev could also become outdated because they change something upstream. Now, it refers at least to what we expect.

storopoli:
The link is hard coded to version 0.15.1 can’t we do stable instead?

lazarusA:
no, Julius is looking into it, it seems there is a problem with the links, all internal ones are relative which is not good. For now, this will work. In the mean time I will keep an eye on it.

storopoli:
Ok so I will unclose the issue

PR #84, opened 2021-08-30
added cup of coffee image :D

lazarusA:


src/makie.jl
```diff
@@ -587,6 +587,9 @@ end
 
 function mixing_surface_contour3d_contour_and_contourf()
     GLMakie.activate!() # hide
+    url = "https://raw.githubusercontent.com/JuliaImages/TestImages.jl/images/images/coffee.png"
```
storopoli, line 590: Is this link going to be static? Isn't it better for us to use a image in our repo? We could add the image if the license permits to our repo.

lazarusA: is static, as far as I know. But, yes, I think we can have it in our site as well... I could do it in the next iteration.

rikhuijzer: Why not
```
using TestImages
img = testimage("coffee.png")
```?

lazarusA: I was under the impression that Images, TestImages is a huge dependency.... not sure if that true nowadays.

lazarusA: But, if you guys are ok with that, I would prefer it this way 😄 , using TestImages is definitely better.

storopoli: @rikhuijzer is the criteria for new dependencies/files. If he is ok with it, them I'm also ok.

rikhuijzer: Yeah sounds good. Adding images will increase repository size (almost) irreversibly, but adding a package can easily be reversed.

PR #93, opened 2021-09-01
Review Julia basics up until arrays

rikhuijzer:
I've renamed it to Julia Basics because my brain was thinking "what does 101 mean?" all the time. I've heard the term before, but I can't seem to remember it.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -4,7 +4,7 @@
   contents = [
     "preface",
     "why_julia",
-    "julia_101",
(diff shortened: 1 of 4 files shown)


storopoli (approved):
This is awesome. Julia 101 is so much US-centric. Definitely something we should change.

storopoli (approved):
This is awesome. Julia 101 is so much US-centric. Definitely something we should change.

PR #104, opened 2021-09-05
Switch to Fathom analytics

rikhuijzer:
Switching to another analytics provider has two benefits:

1. We are GDPR compliant again without having to add a consent button.
2. Less people block Fathom analytics than Google analytics, so the information should be slightly more reliable.

Fathom analytics costs $14/month which I'll pay for now :+1:.


storopoli:
Let’s split the fare?

storopoli (approved):

rikhuijzer:
> Let’s split the fare?

It's okay. It's my hobby :+1:. Alternatively, I would only have spent it on doing burnouts with a moped or something.

PR #107, opened 2021-09-07
Add work in progress mention on homepage

rikhuijzer:
Since Friday, about 130 unique people have looked at the front page. However, the front page doesn't indicate that the book is unfinished which, I expect, can be very frustrating for visitors when they click a link in the menu without content


contents/index.md
```diff
@@ -7,11 +7,31 @@ This file is only included on the website.
 Welcome! This is an open source and open access book on how to do **Data Science using [Julia](https://julialang.org)**.
 Our target audience are PhD candidates from all fields of applied sciences.
 Industry analysts, upcoming and established researchers are also welcome.
+You can navigate through the pages of this website by using the arrow keys (left/right) on your keyboard.
```
lazarusA, line 10:
```suggestion
You can navigate through the pages of e-book by using the arrow keys (left/right) on your keyboard.
```

PR #109, opened 2021-09-08
Set CSV.jl compat to 0.9

rikhuijzer:
I manually compared the local page to `https://juliadatascience.io/load_save` and didn't see any problems.

We should, probably, add tests to automate this manual skimming, but I currently don't think it's worth the time. We just need to keep in mind that we add a test whenever we get issue reports in the future about wrong outputs.


storopoli:
I will cover this in the new performance section. All that `reduce` and `mapreduce` stuff will be outdated then

PR #113, opened 2021-09-09
Add roadmap

rikhuijzer:
Closes #113.
 
@storopoli, I couldn't figure out to add the roadmap nicely to the PDF. Instead, there is now clear message at the bottom of every second page which redirects users to the homepage:

image

Also, each generated PDF now contains the date and time of the build.

PR #114, opened 2021-09-09
Added Performance section to DataFrames chapter

### LINKED ISSUE #100: [DataFrames] Performance
- [x] Allocations: functions with a shebang `!` versus without (e.g., `filter!` versus `filter`)
- [x] Copying vs not copying: `df[!, col]` versus `df[:, col]` ([link](https://dataframes.juliadata.org/dev/man/getting_started/))
- [x] `CSV.jl` reading a lot of `.csv` files with the new version 0.9 and passing a vector of `String`s, details [here](https://github.com/JuliaData/CSV.jl/blob/main/src/file.jl#L65-L67)
- [x] `CSV.File` versus `CSV.read`. Details in the [index of the `CSV.jl` documentation](https://csv.juliadata.org/stable/)
- [x] `CategoricalArrays.jl` compression with `compress=true`

storopoli:
- [x] Allocations: functions with a shebang `!` versus without (e.g., `filter!` versus `filter`)
- [x] Copying vs not copying: `df[!, col]` versus `df[:, col]` ([link](https://dataframes.juliadata.org/dev/man/getting_started/))
- [x] `CSV.jl` reading a lot of `.csv` files with the new version 0.9 and passing a vector of `String`s, details [here](https://github.com/JuliaData/CSV.jl/blob/main/src/file.jl#L65-L67)
- [x] `CSV.File` versus `CSV.read`. Details in the [index of the `CSV.jl` documentation](https://csv.juliadata.org/stable/)
- [x] `CategoricalArrays.jl` compression with `compress=true`

Closes #100

### DIFF
--- a/contents/dataframes_performance.md
+++ b/contents/dataframes_performance.md
@@ -6,18 +6,18 @@ In this section, we will give some performance tips and tricks.
 
 ### Allocations
 
-Like we explained in @sec:function_bang functions that end with a bang `!` are a common pattern to denote functions that modify one or more of their arguments.
+Like we explained in @sec:function_bang, functions that end with a bang `!` are a common pattern to denote functions that modify one or more of their arguments.
 In the context of high performance Julia code, this *means* that **functions with `!` do not perform any sort of allocation**.
-They just change in place any object that we supply as arguments.
+They just change in-place one of the objects that we supply as arguments.
 
 Almost all the `DataFrames.jl` functions that we've seen so far have a \"`!` twin\".
 For example, `filter` has a non allocating `filter!`, `select` has `select!`, `subset` has `subset!`, and so on.
-Notice that these functions **do not** return a `DataFrame`, but instead they **alter** the `DataFrame` that they act upon.
+Notice that these functions **do not** return a new `DataFrame`, but instead they **alter** the `DataFrame` that they act upon.
 Additionally, there are some functions that do not have a `!` counterpart.
-For example, all the `join`s, by logical purposes, *cannot* be done without allocations.
+For example, all the `join`s, for technical reasons, *cannot* be done in-place.
 Thus, we don't have any `join!` functions in `DataFrames.jl`.
 
-If you want the highest speed and performance in your code you should definitely use the `!` functions instead of regular `DataFrames.jl` functions.
+If you want the highest speed and performance in your code, you should definitely use the `!` functions instead of regular `DataFrames.jl` functions.
 
 Let's go back to the example of the `select` function in the beginning of @sec:select.
 Here is the responses `DataFrame`:
@@ -66,7 +66,8 @@ s = """
 sco(s; process=string, post=plainblock)
 ```
 
-As we can see `select!` makes less allocations that `select` so it should be faster while consuming less memory.
+As we can see, `select!` allocates less than `select`.
+So, it should be faster and while consuming less memory.
 
 ### Copying vs Not Copying Columns
 
@@ -107,10 +108,10 @@ So, if you don't need a copy always access your `DataFrame`s columns with `df.co
 
 ### CSV.read versus CSV.File
 
-If you take a look at the help output for `CSV.read` you will see that it is a convenience function identical to a function called `CSV.File` with the same keywords arguments.
+If you take a look at the help output for `CSV.read`, you will see that it is a convenience function identical to the function called `CSV.File` with the same keywords arguments.
 Both `CSV.read` and `CSV.File` will read the contents of a CSV file, but they differ in the default behavior.
 **`CSV.read`, by default, will not make copies** of the incoming data.
-Instead, `CSV.read` will pass all the data to the second argument (known as "sink").
+Instead, `CSV.read` will pass all the data to the second argument (known as the "sink").
 
 So, something like this:
 
@@ -120,8 +121,9 @@ df = CSV.read("file.csv", DataFrame)
 
 will pass all the incoming data from `file.csv` to the `DataFrame` sink, thus returning a `DataFrame` type that we store in the `df` variable.
 
-For the case of **`CSV.File` the default behavior is the opposite: it will make copies of every column contained in the CSV file**.
-Also the syntax is slight different, we need to wrap anything that `CSV.File` returns in a `DataFrame` constructor function:
+For the case of **`CSV.File`, the default behavior is the opposite: it will make copies of every column contained in the CSV file**.
+Also, the syntax is slightly different.
+We need to wrap anything that `CSV.File` returns in a `DataFrame` constructor function:
 
 ```julia
 df = DataFrame(CSV.File("file.csv"))
@@ -160,7 +162,6 @@ files = filter(endswith(".csv"), readdir())
 df = CSV.read(files, DataFrame)
 ```
 
-Much more simpler.
 `CSV.jl` will design a file for each thread available in the computer while it lazily concatenates each thread parsed output into a `DataFrame`.
 So we have the **additional benefit of multithreading** that we don't have with the `reduce` option.
 
@@ -183,7 +184,7 @@ That's why `categorical` has a `compress` argument that accepts either `true` or
 If you pass **`compress=true`, `CategoricalArrays.jl` will try to compress the underlying categorical to the smallest possible representation in `UInt`**.
 For example, the previous `categorical` vector would be represented as an unsigned integer of size 8 bits `UInt8` (mostly because this is the smallest unsigned integer available in Julia):
 
-[^bigdata]: also notice that regular data is not big data. So if you are dealing primarily with big data please take caution in capping your categorical values.
+[^bigdata]: also notice that regular data (about <10 000 rows) is not big data (about >100 000 rows). So, if you are dealing primarily with big data please take caution in capping your categorical values.
 
 ```jl
 s = """
@@ -192,9 +193,9 @@ s = """
 sco(s; process=string, post=plainblock)
 ```
 
-What does this all means?
+What does this all mean?
 Suppose you have a big vector.
-For example a vector with one million entries, but only 4 underlying categories: A, B, C or D.
+For example, a vector with one million entries, but only 4 underlying categories: A, B, C or D.
 If you do not compress the resulting categorical vector, you will have one million entries stored as `UInt32`.
 On the other hand, if you do compress it, you will have one million entries stored instead as `UInt8`.
 By using `Base.summarysize` function we can get the underlying size, in bytes, of a given object.
--- a/contents/dataframes_performance.md
+++ b/contents/dataframes_performance.md
@@ -4,15 +4,14 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
 Like everything in Julia, `DataFrames.jl` can be really fast.
 In this section, we will give some performance tips and tricks.
 
-### Allocations
+### In-place operations 
 
 Like we explained in @sec:function_bang, functions that end with a bang `!` are a common pattern to denote functions that modify one or more of their arguments.
-In the context of high performance Julia code, this *means* that **functions with `!` do not perform any sort of allocation**.
-They just change in-place one of the objects that we supply as arguments.
+In the context of high performance Julia code, this *means* that **functions with `!` will just change in-place the objects that we have supplied as arguments.
 
-Almost all the `DataFrames.jl` functions that we've seen so far have a \"`!` twin\".
-For example, `filter` has a non allocating `filter!`, `select` has `select!`, `subset` has `subset!`, and so on.
-Notice that these functions **do not** return a new `DataFrame`, but instead they **alter** the `DataFrame` that they act upon.
+Almost all the `DataFrames.jl` functions that we've seen have a \"`!` twin\".
+For example, `filter` has an _in-place_ `filter!`, `select` has `select!`, `subset` has `subset!`, and so on.
+Notice that these functions **do not** return a new `DataFrame`, but instead they **update** the `DataFrame` that they act upon.
 Additionally, there are some functions that do not have a `!` counterpart.
 For example, all the `join`s, for technical reasons, *cannot* be done in-place.
 Thus, we don't have any `join!` functions in `DataFrames.jl`.
@@ -36,7 +35,7 @@ s = """
 sco(s, process=without_caption_label)
 ```
 
-And here is the non allocating function:
+And here is the _in-place_ function:
 
 ```jl
 s = """
@@ -47,7 +46,7 @@ sco(s; process=without_caption_label)
 ```
 
 The `@allocated` macro tells us how much memory was allocated.
-In other words, how much new information the computer had to store in its memory while running the code.
+In other words, **how much new information the computer had to store in its memory while running the code**.
 Let's see how they will perform: 
 
 ```jl
@@ -84,7 +83,7 @@ Additionally, the syntax `df.col` is the same as `df[!, :col]` with the bang `!`
 The second way to access a `DataFrame` column is the `df[:, :col]` with the colon `:` as the row selector.
 This kind of access **does copy** the column `col`, so beware that it may produce unwanted allocations.
 
-As before, let's try out our two ways to access a column in the responses `DataFrame`:
+As before, let's try out these two ways to access a column in the responses `DataFrame`:
 
 ```jl
 s = """
@@ -110,7 +109,7 @@ So, if you don't need a copy always access your `DataFrame`s columns with `df.co
 
 ### CSV.read versus CSV.File
 
-If you take a look at the help output for `CSV.read`, you will see that it is a convenience function identical to the function called `CSV.File` with the same keywords arguments.
+If you take a look at the help output for `CSV.read`, you will see that there is a convenience function identical to the function called `CSV.File` with the same keywords arguments.
 Both `CSV.read` and `CSV.File` will read the contents of a CSV file, but they differ in the default behavior.
 **`CSV.read`, by default, will not make copies** of the incoming data.
 Instead, `CSV.read` will pass all the data to the second argument (known as the "sink").
@@ -148,16 +147,16 @@ Now let's turn our attention to the `CSV.jl`.
 Specially when we have multiple CSV files to read into a single `DataFrame`.
 Since version 0.9 of `CSV.jl` we can provide a vector of strings representing filenames.
 Before, we needed to perform some sort of multiple file reading and then concatenate vertically the results into a single `DataFrame`.
-To exemplify, the code below reads from multiple CSV files and them concatenates vertically using `vcat` them into a single `DataFrame` with the `reduce` function:
+To exemplify, the code below reads from multiple CSV files and then concatenates them vertically using `vcat` into a single `DataFrame` with the `reduce` function:
 
 ```julia
 files = filter(endswith(".csv"), readdir())
 df = reduce(vcat, CSV.read(file, DataFrame) for file in files)
 ```
 
-One additional quirk is that `reduce` will not paralelize because it needs to keep the order of `vcat` which follows the same ordering of the `files` vector.
+One additional trait is that `reduce` will not paralelize because it needs to keep the order of `vcat` which follows the same ordering of the `files` vector.
 
-With the new functionality in `CSV.jl` we simply pass the `files` vector into the `CSV.read` function:
+With this functionality in `CSV.jl` we simply pass the `files` vector into the `CSV.read` function:
 
 ```julia
 files = filter(endswith(".csv"), readdir())
@@ -171,7 +170,7 @@ So we have the **additional benefit of multithreading** that we don't have with 
 
 If you are handling data with a lot of categorical values, i.e. a lot of columns with textual data that represent somehow different qualitative data, you would probably benefit by using `CategoricalArrays.jl` compression.
 
-By default, **`CategoricalArrays.jl` will use a unsigned integer of size 32 bits `UInt32` to represent the underlying categories**:
+By default, **`CategoricalArrays.jl` will use an unsigned integer of size 32 bits `UInt32` to represent the underlying categories**:
 
 ```jl
 s = """
--- a/contents/dataframes_performance.md
+++ b/contents/dataframes_performance.md
@@ -46,7 +46,9 @@ s = """
 sco(s; process=without_caption_label)
 ```
 
-Now let's see how they perform with the `@allocated` macro:
+The `@allocated` macro tells us how much memory was allocated.
+In other words, how much new information the computer had to store in its memory while running the code.
+Let's see how they will perform: 
 
 ```jl
 s = """
--- a/contents/dataframes_performance.md
+++ b/contents/dataframes_performance.md
@@ -4,7 +4,7 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
 Like everything in Julia, `DataFrames.jl` can be really fast.
 In this section, we will give some performance tips and tricks.
 
-### In-place operations 
+### In-place operations {#sec:df_performance_inplace}
 
 Like we explained in @sec:function_bang, functions that end with a bang `!` are a common pattern to denote functions that modify one or more of their arguments.
 In the context of high performance Julia code, this *means* that **functions with `!` will just change in-place the objects that we have supplied as arguments.
@@ -70,7 +70,7 @@ sco(s; process=string, post=plainblock)
 As we can see, `select!` allocates less than `select`.
 So, it should be faster and while consuming less memory.
 
-### Copying vs Not Copying Columns
+### Copying vs Not Copying Columns {#sec:df_performance_df_copy}
 
 There are **two ways to access a DataFrame column**.
 They differ in how they are accessed: one creates a "view" to the column without copying and the other creates a whole new column by copying the original column.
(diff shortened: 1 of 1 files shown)


contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
 Like everything in Julia, `DataFrames.jl` can be really fast.
 In this section, we will give some performance tips and tricks.
 
-```{=comment}
-- Functions with a shebang versus without (e.g., filter! versus filter)
-- df[!, col] versus df[:, col]
-- other things? Jose probably knows some more
+### Allocations
+
+Like we explained in @sec:function_bang functions that end with a bang `!` are a common pattern to denote functions that modify one or more of their arguments.
```
rikhuijzer, line 9:
```suggestion
Like we explained in @sec:function_bang, functions that end with a bang `!` are a common pattern to denote functions that modify one or more of their arguments.
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+In the context of high performance Julia code, this *means* that **functions with `!` do not perform any sort of allocation**.
```
rikhuijzer, line 10: I don't that this holds in general. As a counterexample, I can think of a transform operation with some complex logic inside the transformation function.

```
julia> df = DataFrame(A = [1]);

julia> @allocated transform!(df, :A => x -> (rand(100); x))
1168837
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+They just change in place any object that we supply as arguments.
```
rikhuijzer, line 11:
```suggestion
They just change in-place one of the objects that we supply as arguments.
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+
+Almost all the `DataFrames.jl` functions that we've seen so far have a \"`!` twin\".
+For example, `filter` has a non allocating `filter!`, `select` has `select!`, `subset` has `subset!`, and so on.
+Notice that these functions **do not** return a `DataFrame`, but instead they **alter** the `DataFrame` that they act upon.
```
rikhuijzer, line 15:
```suggestion
Notice that these functions **do not** return a new `DataFrame`, but instead they **alter** the `DataFrame` that they act upon.
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+Additionally, there are some functions that do not have a `!` counterpart.
+For example, all the `join`s, by logical purposes, *cannot* be done without allocations.
```
rikhuijzer, line 17:
```suggestion
For example, all the `join`s, for technical reasons, *cannot* be done in place.
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+Thus, we don't have any `join!` functions in `DataFrames.jl`.
+
+If you want the highest speed and performance in your code you should definitely use the `!` functions instead of regular `DataFrames.jl` functions.
```
rikhuijzer, line 20:
```suggestion
If you want the highest speed and performance in your code, you should definitely use the `!` functions instead of regular `DataFrames.jl` functions.
```
"You should definitely use [...]" is a sentence which can stand on its own. Therefore, there should be a comma.

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+
+```jl
+s = """
+    # non allocarting function # hide
+    select!(responses(), :id, :q1)
+    """
+sco(s; process=without_caption_label)
+```
+
+Now let's see how they perform with the `@allocated` macro:
```
rikhuijzer, line 49: Maybe, add
"The allocated macro tells us how much memory was allocated. In other words, how much new information the computer had to store in its memory while running the code."

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+```jl
+s = """
+    # non allocation # hide
+    df = responses()
+    @allocated select!(df, :id, :q1)
+    """
+sco(s; process=string, post=plainblock)
+```
+
+As we can see `select!` makes less allocations that `select` so it should be faster while consuming less memory.
```
rikhuijzer, line 69:
```suggestion
As we can see, `select!` allocates less than `select`.
So, it should be faster and while consuming less memory.
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+
+### Copying vs Not Copying Columns
+
+There are **two ways to access a DataFrame column**.
+They differ in how they are accessed: one creates a "view" to the column without copying and the other creates a whole new column by copying the original column.
+
+The first is the regular dot `.` operator followed by the column name, like in `df.Col`.
```
rikhuijzer, line 76:
```suggestion
The first is the regular dot `.` operator followed by the column name, like in `df.col`.
```
Below, we used lowercase columns, so maybe here too?

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+s = """
+    # non allocation # hide
+    df = responses()
+    @allocated col = df[!, :id]
+    """
+sco(s; process=string, post=plainblock)
+```
+
+When we access a column without copying it we are making zero allocations and our code should be faster.
+So, if you don't need a copy always access your `DataFrame`s columns with `df.col` or `df[!, :col]` instead of `df[:, :col]`.
```
rikhuijzer, line 106: Nice! :rocket: I will go back to this section very often, I think

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+
+### CSV.read versus CSV.File
+
+If you take a look at the help output for `CSV.read` you will see that it is a convenience function identical to a function called `CSV.File` with the same keywords arguments.
```
rikhuijzer, line 110:
```suggestion
If you take a look at the help output for `CSV.read`, you will see that it is a convenience function identical to the function called `CSV.File` with the same keywords arguments.
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+Both `CSV.read` and `CSV.File` will read the contents of a CSV file, but they differ in the default behavior.
+**`CSV.read`, by default, will not make copies** of the incoming data.
+Instead, `CSV.read` will pass all the data to the second argument (known as "sink").
```
rikhuijzer, line 113:
```suggestion
Instead, `CSV.read` will pass all the data to the second argument (known as the "sink").
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+
+So, something like this:
+
+```julia
+df = CSV.read("file.csv", DataFrame)
+```
+
+will pass all the incoming data from `file.csv` to the `DataFrame` sink, thus returning a `DataFrame` type that we store in the `df` variable.
+
+For the case of **`CSV.File` the default behavior is the opposite: it will make copies of every column contained in the CSV file**.
```
rikhuijzer, line 123:
```suggestion
For the case of **`CSV.File`, the default behavior is the opposite: it will make copies of every column contained in the CSV file**.
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+Also the syntax is slight different, we need to wrap anything that `CSV.File` returns in a `DataFrame` constructor function:
```
rikhuijzer, line 124:
```suggestion
Also, the syntax is slightly different.
We need to wrap anything that `CSV.File` returns in a `DataFrame` constructor function:
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+
+```julia
+df = DataFrame(CSV.File("file.csv"))
+```
+
+Or, in a more preferred idiomatic syntax with the pipe `|>` operator:
```
rikhuijzer, line 130: I remember a discussion a while back where Tamas Papp said that almost in all cases
```
a = f(x)
b = g(a)
```
is better than
```
a = g(f(x))
```
or
```
a = f(x) |> g
```
The reason is pretty straightforward: if you want to debug the first way, you can easily pin the debugger to a specifc point in the operation. Also, you can easily comment out or remove a part of the operation and it's more clear because the intermediate variable can get a clear name.

That said, I don't think it's wrong per se and am curious for your thoughts. I love the pipe operator when I'm doing stuff in the REPL. If you prefer the pipe operator here, then that's fine.

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+
+
+```julia
+df = CSV.File("file.csv") |> DataFrame
+```
+
+Like we said, `CSV.File` will make copies of each column in the underlying CSV file.
+Ultimately, if you want the most performance, you would definitely use `CSV.read` instead of `CSV.File`.
+That's why we only covered `CSV.read` in @sec:csv.
```
rikhuijzer, line 139: yeeeeaaahhh for sure :stuck_out_tongue: exactly what I had in mind! :stuck_out_tongue: you expect too much of me :stuck_out_tongue:

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+One additional quirk is that `reduce` will not paralelize because it needs to keep the order of `vcat` which follows the same ordering of the `files` vector.
+
+With the new functionality in `CSV.jl` we simply pass the `files` vector into the `CSV.read` function:
+
+```julia
+files = filter(endswith(".csv"), readdir())
+df = CSV.read(files, DataFrame)
+```
+
+Much more simpler.
```
rikhuijzer, line 163:
```suggestion
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+df = CSV.File("file.csv") |> DataFrame
+```
+
+Like we said, `CSV.File` will make copies of each column in the underlying CSV file.
+Ultimately, if you want the most performance, you would definitely use `CSV.read` instead of `CSV.File`.
+That's why we only covered `CSV.read` in @sec:csv.
+
+### CSV.jl Multiple Files
+
+Now let's turn our attention to the `CSV.jl`.
```
rikhuijzer, line 143: Super cool section! :rocket:

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+sco(s; process=string, post=plainblock)
+```
+
+This means that `CategoricalArrays.jl` can represent up to $2^{32}$ different categories in a given vector or column, which is a huge value (close to 4.3 billion).
+You probably would never need to have this sort of capacity in dealing with regular data[^bigdata].
+That's why `categorical` has a `compress` argument that accepts either `true` or `false` to whether compress or not the underlying categorical data.
+If you pass **`compress=true`, `CategoricalArrays.jl` will try to compress the underlying categorical to the smallest possible representation in `UInt`**.
+For example, the previous `categorical` vector would be represented as an unsigned integer of size 8 bits `UInt8` (mostly because this is the smallest unsigned integer available in Julia):
+
+[^bigdata]: also notice that regular data is not big data. So if you are dealing primarily with big data please take caution in capping your categorical values.
```
rikhuijzer, line 186:
```suggestion
[^bigdata]: also notice that regular data (about <10 000 rows) is not big data (about >100 000 rows). So, if you are dealing primarily with big data please take caution in capping your categorical values.
```
Maybe?

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+
+```jl
+s = """
+    typeof(categorical(["A", "B", "C"]; compress=true))
+    """
+sco(s; process=string, post=plainblock)
+```
+
+What does this all means?
```
rikhuijzer, line 195:
```suggestion
What does this all mean?
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+Suppose you have a big vector.
+For example a vector with one million entries, but only 4 underlying categories: A, B, C or D.
```
rikhuijzer, line 197:
```suggestion
For example, a vector with one million entries, but only 4 underlying categories: A, B, C or D.
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,237 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+So let's quantify how much more memory we would need to have if we did not compressed our one million categorical vector:
+
+```julia
+using Random 
+```
+
+```jl
+s = """
+    one_mi_vec = rand(["A", "B", "C", "D"], 1_000_000)
+    Base.summarysize(categorical(one_mi_vec))
```
rikhuijzer, line 210: Very nice :+1: :+1:

rikhuijzer (commented):
Interesting read!

rikhuijzer:
I've already marked the DataFrames section as done on the homepage in #113.

contents/dataframes_performance.md
storopoli: How do you advise a better phrasing?

contents/dataframes_performance.md
storopoli: We are using in-place.

contents/dataframes_performance.md
storopoli: Actually we need to change it all then.

contents/dataframes_performance.md
storopoli: Hahahaha

contents/dataframes_performance.md
rikhuijzer: Remove the sentence and only mention the in place part.

contents/dataframes_performance.md
storopoli: If I do JuliaFormatter.jl to the file it fixes all the pipe stuff. This:

`f(x) |> g`

Becomes:

`g(f(x))`

rikhuijzer: JuliaFormatter isn't perfect either. It also has some non-idiomatic defaults

contents/dataframes_performance.md
storopoli: Thanks, I've added it.

storopoli:
Ready for another round. I would also love to get @lazarusA opinions.

contents/dataframes_performance.md
storopoli: Can you suggest in the PR so I can see it and apply or edit if necessary?

contents/dataframes_performance.md
storopoli: I use JuliaFormatter blue code style.
https://domluna.github.io/JuliaFormatter.jl/dev/blue_style/

rikhuijzer (approved):

contents/dataframes_performance.md
```diff
@@ -4,8 +4,240 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
 In this section, we will give some performance tips and tricks.
 
-```{=comment}
-- Functions with a shebang versus without (e.g., filter! versus filter)
-- df[!, col] versus df[:, col]
-- other things? Jose probably knows some more
+### Allocations
+
+Like we explained in @sec:function_bang, functions that end with a bang `!` are a common pattern to denote functions that modify one or more of their arguments.
+In the context of high performance Julia code, this *means* that **functions with `!` do not perform any sort of allocation**.
```
lazarusA, line 10:
```suggestion
In the context of high performance Julia code, this *means* that **functions with `!` will just change in-place the objects that we have supplied as arguments.
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,240 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+They just change in-place one of the objects that we supply as arguments.
```
lazarusA, line 11:
```suggestion
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,240 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+
+Almost all the `DataFrames.jl` functions that we've seen so far have a \"`!` twin\".
```
lazarusA, line 13:
```suggestion
Almost all the `DataFrames.jl` functions that we've seen have a \"`!` twin\".
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,240 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+For example, `filter` has a non allocating `filter!`, `select` has `select!`, `subset` has `subset!`, and so on.
```
lazarusA, line 14:
```suggestion
For example, `filter` has an _in-place_ `filter!`, `select` has `select!`, `subset` has `subset!`, and so on.
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,240 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
 Like everything in Julia, `DataFrames.jl` can be really fast.
 In this section, we will give some performance tips and tricks.
 
-```{=comment}
-- Functions with a shebang versus without (e.g., filter! versus filter)
-- df[!, col] versus df[:, col]
-- other things? Jose probably knows some more
+### Allocations
```
lazarusA, line 7:
```suggestion
### In-place operations 
```

lazarusA: The difference between having or not having allocations is a little more than just using a bang !

contents/dataframes_performance.md
```diff
@@ -4,8 +4,240 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+
+Like we explained in @sec:function_bang, functions that end with a bang `!` are a common pattern to denote functions that modify one or more of their arguments.
+In the context of high performance Julia code, this *means* that **functions with `!` do not perform any sort of allocation**.
+They just change in-place one of the objects that we supply as arguments.
+
+Almost all the `DataFrames.jl` functions that we've seen so far have a \"`!` twin\".
+For example, `filter` has a non allocating `filter!`, `select` has `select!`, `subset` has `subset!`, and so on.
+Notice that these functions **do not** return a new `DataFrame`, but instead they **alter** the `DataFrame` that they act upon.
```
lazarusA, line 15:
```suggestion
Notice that these functions **do not** return a new `DataFrame`, but instead they **update** the `DataFrame` that they act upon.
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,240 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+
+```jl
+s = """
+    # allocating function # hide
+    select(responses(), :id, :q1)
+    """
+sco(s, process=without_caption_label)
+```
+
+And here is the non allocating function:
```
lazarusA, line 39:
```suggestion
And here is the _in-place_ function:
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,240 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+
+The first is the regular dot `.` operator followed by the column name, like in `df.col`.
+This kind of access **does not copy** the column `col`.
+Instead `df.col` creates a "view" which is a link to the original column without performing any allocation.
+Additionally, the syntax `df.col` is the same as `df[!, :col]` with the bang `!` as the row selector.
+
+The second way to access a `DataFrame` column is the `df[:, :col]` with the colon `:` as the row selector.
+This kind of access **does copy** the column `col`, so beware that it may produce unwanted allocations.
+
+As before, let's try out our two ways to access a column in the responses `DataFrame`:
```
lazarusA, line 87:
```suggestion
As before, let's try out these two ways to access a column in the responses `DataFrame`:
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,240 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+```jl
+s = """
+    # non allocarting function # hide
+    select!(responses(), :id, :q1)
+    """
+sco(s; process=without_caption_label)
+```
+
+The `@allocated` macro tells us how much memory was allocated.
+In other words, how much new information the computer had to store in its memory while running the code.
```
lazarusA, line 50:
```suggestion
In other words, **how much new information the computer had to store in its memory while running the code**.
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,240 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+    """
+sco(s; process=string, post=plainblock)
+```
+
+When we access a column without copying it we are making zero allocations and our code should be faster.
+So, if you don't need a copy always access your `DataFrame`s columns with `df.col` or `df[!, :col]` instead of `df[:, :col]`.
+
+### CSV.read versus CSV.File
+
+If you take a look at the help output for `CSV.read`, you will see that it is a convenience function identical to the function called `CSV.File` with the same keywords arguments.
```
lazarusA, line 113:
```suggestion
If you take a look at the help output for `CSV.read`, you will see that there is a convenience function identical to the function called `CSV.File` with the same keywords arguments.
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,240 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+**`CSV.read`, by default, will not make copies** of the incoming data.
+Instead, `CSV.read` will pass all the data to the second argument (known as the "sink").
+
+So, something like this:
+
+```julia
+df = CSV.read("file.csv", DataFrame)
+```
+
+will pass all the incoming data from `file.csv` to the `DataFrame` sink, thus returning a `DataFrame` type that we store in the `df` variable.
```
lazarusA, line 124: is this 100% true? when doing the passing is not doing the copying to the DataFrame as well?

contents/dataframes_performance.md
```diff
@@ -4,8 +4,240 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+Ultimately, if you want the most performance, you would definitely use `CSV.read` instead of `CSV.File`.
+That's why we only covered `CSV.read` in @sec:csv.
+
+### CSV.jl Multiple Files
+
+Now let's turn our attention to the `CSV.jl`.
+Specially when we have multiple CSV files to read into a single `DataFrame`.
+Since version 0.9 of `CSV.jl` we can provide a vector of strings representing filenames.
+Before, we needed to perform some sort of multiple file reading and then concatenate vertically the results into a single `DataFrame`.
+To exemplify, the code below reads from multiple CSV files and them concatenates vertically using `vcat` them into a single `DataFrame` with the `reduce` function:
```
lazarusA, line 151:
```suggestion
To exemplify, the code below reads from multiple CSV files and then concatenates them vertically using `vcat` into a single `DataFrame` with the `reduce` function:
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,240 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+
+```julia
+files = filter(endswith(".csv"), readdir())
+df = reduce(vcat, CSV.read(file, DataFrame) for file in files)
+```
+
+One additional quirk is that `reduce` will not paralelize because it needs to keep the order of `vcat` which follows the same ordering of the `files` vector.
```
lazarusA, line 158:
```suggestion
One additional trait is that `reduce` will not paralelize because it needs to keep the order of `vcat` which follows the same ordering of the `files` vector.
```

contents/dataframes_performance.md
```diff
@@ -4,8 +4,240 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+
+With the new functionality in `CSV.jl` we simply pass the `files` vector into the `CSV.read` function:
```
lazarusA, line 160:
```suggestion
With this functionality in `CSV.jl` we simply pass the `files` vector into the `CSV.read` function:
```
soon it will not be new, just the norm.

contents/dataframes_performance.md
```diff
@@ -4,8 +4,240 @@ So far, we didn't think about making our `DataFrames.jl` code **fast**.
...
+```
+
+`CSV.jl` will design a file for each thread available in the computer while it lazily concatenates each thread parsed output into a `DataFrame`.
+So we have the **additional benefit of multithreading** that we don't have with the `reduce` option.
+
+### CategoricalArrays.jl compression
+
+If you are handling data with a lot of categorical values, i.e. a lot of columns with textual data that represent somehow different qualitative data, you would probably benefit by using `CategoricalArrays.jl` compression.
+
+By default, **`CategoricalArrays.jl` will use a unsigned integer of size 32 bits `UInt32` to represent the underlying categories**:
```
lazarusA, line 174:
```suggestion
By default, **`CategoricalArrays.jl` will use an unsigned integer of size 32 bits `UInt32` to represent the underlying categories**:
```

lazarusA (commented):
Just some small comments, in general it looks great.

contents/dataframes_performance.md
storopoli: Yes taken strictly from the CSV.jl documentation and some discourse threads

storopoli:
Thanks guys for the feedback. I've just added pandoc section ids that were missing the `{#sec:foo}` stuff.

PR #116, opened 2021-09-12
Resize simple image

rikhuijzer:
In response to https://github.com/rikhuijzer/Books.jl/issues/192. I've set the width to 70% now for simple images.

Note that this will change the width in both the HTML and PDF output. This makes it easier when developing with HTML to imagine how it is going to look in the PDF.


lazarusA:
This is a really nice feature.

Issue #112, opened 2021-09-09
Statistics Chapter


storopoli:
A final chapter with basic statistics using both `DataFrames.jl` and `Makie.jl`. Each concept we should give the intuition behind it and also get into mathematical details.

We should cover:

- [x] Statistics: Quick Intro and Importance
- [x] Central Tendencies Measures: mean, median and mode
- [x] Dispersion Measures: standard deviation, variance, mean absolute deviation, quartile and IQR
- [x] Dependence Measures: Correlation and Covariance
- [x] Distributions
	- [x] Normal vs Non-Normal
	- [x] Discrete vs Continuous
	- [x] PDF
	- [x] CDF
- [x] Statistical Visualizations
	- [x] Box plots
	- [x] Histograms
	- [x] Density plots
	- [x] Anscombe Quartet

AriMKatz:
Hi,

This is a great book. I'm not sure about license compatibility or if there's room for collaboration, but I wanted to point you to: https://github.com/unbalancedparentheses/data_science_in_julia_for_hackers 

Which takes a bayesian approach to introducing stats. Also has ML, sciml and TS forecasting chapters.

storopoli:
@AriMKatz DS in Julia for Hackers is am amazing resource. I would like to hear @rikhuijzer and @lazarusA  opinions also.

PR #119, opened 2021-09-15
First Section of Statistics Chapter

storopoli:
This section covers
- [x] Statistics: Quick Intro and Importance
- [x] Central Tendencies Measures: mean, median and mode
- [x] Dispersion Measures: standard deviation, variance, mean absolute deviation, quartile and IQR
- [x] Dependence Measures: Correlation and Covariance

Partially deals with #112


contents/stats.md
```diff
@@ -0,0 +1,445 @@
+# Statistics {#sec:stats}
+
+This content was created in order to awaken the reader to the importance of statistics for science and knowledge generation.
```
rikhuijzer, line 3:
```suggestion
An important part of data science is being able to make predictions.
In essence, the world is constantly generating data and we're only seeing the small part of the data that we captured.
This means that we have to take this small part and infer things about the real world.
This is called statistics.
```
Something like this? I avoid "awaken" because I don't think that a typical reader of our book would not already know the importance of statistics.

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+* measures of **central tendencies**
+* measures of **dispersion**
+* measures of **dependence**
+* **probability distributions**
+* statistical **visualization**
+
+Statistics is important because it is a **tool to make sense of data**.
+With the abundant availability of data, we are often overwhelmed by numbers.
+Statistics offers a way to comprehend, summarize and infer information from data.
+We believe that every data scientist should have a basic understanding of statistics and how to perform simple statistical operations.
```
rikhuijzer, line 19:
```suggestion
We believe that a basic understanding of statistics and how to perform simple statistical operations are valuable skills for every data scientist
```
"should" sounds a bit forced

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+
+We can divide statistics into two broad categories: **descriptive** and **inferential**.
+**Descriptive statistics summarizes and quantifies the characteristics of a given observed data**.
+Common metrics are: mean, median, mode, standard deviation, variance, correlation, percentiles.
```
rikhuijzer, line 23:
```suggestion
Common descriptives are: mean, median, mode, standard deviation, variance, correlation and percentile.
```

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+
+**Inferential statistics allows generating inferences** (statements) **from observed data about the data generation process**.
+All phenomena have an underlying data generating process that describes how the data is being generated.
+For example, in a soccer game, a scored goal can be explained by an underlying process: a tactic, error, struck of luck; or a mix of those.
+If we know a phenomenon's data generating process, we can use probability to simulate possible scenarios given certain aspects.
+Most of the time, specially in applied sciences, we do not have full knowledge of the data generating process.
```
rikhuijzer, line 29:
```suggestion
Most of the time, especially in applied sciences, we do not have full knowledge of the data generating process.
```

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+Given the observed data, we can retrace our way back to the data generating process.
+This process is known as **statistical inference**.
+Given some data, we can *infer* what are the aspects of the underlying data generating process.
+This is the realm of inferential statistics.
+
+In @fig:stats_vs_prob we summarize the relationship between data generating process and observed data.
```
rikhuijzer, line 35: As a reader with a fresh mind, I would find the image useful a bit earlier (somewhere in the previous paragraph)

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+    fig = statistics_graph()
+    caption = "Statistics vs Probability"
+    label = "stats_vs_prob"
+    Options(fig; filename=label, caption, label)
+    """
+sco(s)
+```
+
+In this chapter, we will cover only descriptive statistics.
+Inferential statistics is an important and fundamental component of applied sciences, but its scope is too broad. Let's begin on simple ways to summarize our data with central tendencies.
```
rikhuijzer, line 50: > Inferential statistics is an important and fundamental component of applied sciences, but its scope is too broad.

I fully understand your point and that we don't want the chapter to be too big. However, the argument seems a bit weak. Explaining a few HypothesisTests.jl shouldn't be too bad, right? For now, I'm also okay without. I can also pick it up later

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+
+## Central Tendencies Measures {#sec:stats_central}
+
+The most basic way of using descriptive statistics is to summarize data by a measure of **central tendency**.
+
+### Mean {#sec:stats_central_mean}
+
+We have several ways to summarize data but the most common is to use the mean.
```
rikhuijzer, line 58:
```suggestion
The most common central tendency measure is the mean.
```
This links better with the previous sentence.

rikhuijzer (commented):
Already a few comments. I'll continue the review tomorrow.

contents/stats.md
storopoli: Statistics is not only predictions. We need to say more about "inference" since our audience are PhD candidates from all fields of applied sciences. That's why I've inserted "importance of statistics for science and knowledge generation"

contents/stats.md
storopoli: Paraphrasing @lazarusA who shook me with a deep truth last meeting: "Well, you are in the wrong field". Do you really believe that "should" sounds forced for our target audience?

contents/stats.md
storopoli: They are still metrics. We can use them for inference. I don't think they are descriptives. Maybe common "descriptive metrics"?

My point is that we can have all of those metrics in a parameter estimation setting for example.

contents/stats.md
storopoli: I don't think we should do this for v1 or 1st edition. But we definitely should incorporate this in v2 and 2nd edition.

contents/stats.md
storopoli: Let's hear @lazarusA's opinion also...

contents/stats.md
rikhuijzer: In essence, when you infer a model from a dataset, you're gonna use that model to make predictions. I'll leave it up to you what you decide to write here. :+1:

contents/stats.md
rikhuijzer: After a bit more thinking. I think it is personal preference. I hate the "you should". "you should" sentences are basically always bad

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+
+```jl
+sco("more_grades()"; process=without_caption_label)
+```
+
+```jl
+# inside stats content # hide
+s = """
+    gdf = groupby(more_grades(), :name)
+    combine(gdf, :grade => mean)
```
rikhuijzer, line 92: This is actually very nice. I've never seen the mean shown via a groupby :100: Fits also very nice together with the rest of the book

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+```
+
+### Median {#sec:stats_central_median}
+
+We will see that the mean is highly sensitive to outliers and can sometimes misguide us, specially when we are dealing with long-tailed (non-normal distributions) (more in @sec:stats_dist_normal).
+That is why sometimes we are interested in also the **median** which is the **middle value that separates the higher half from the lower half of the data**.
+Intuitively, the median tells us the value of the data's 50\% percentile.
+One example is when we are analyzing income.
+The median is the upper limit that we expect that half of the observations earn.
+So, if the median income is U$ 80,000, we expect that half of our observations earn between the minimum value and U$ 80,000.
```
rikhuijzer, line 104:
```suggestion
So, if the median income is $ 80,000, we expect that half of our observations earn between the minimum value and $ 80,000.
```
I've never seen the `U$` notation before. Doesn't really matter for the example

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+The mathematical formula for the median is:
+
+$$ \operatorname{median}(\mathbf{x}) = \frac{x_{\lfloor (\# x+1) \div 2 \rfloor} + x_{\lceil (\# x+1) \div 2 \rceil}}{2}, $$ {#eq:median}
+
+where:
+
+- $\mathbf{x} = x_1, \cdots, x_n$ is an ordered list of numbers
```
rikhuijzer, line 111:
```suggestion
- $\mathbf{x} = x_1, \cdots, x_n$ is an ordered vector of numbers
- $\mathbf{x}_i$ is the element in vector $\mathbf{x}$ at position $i$
```
I propose vector because it's more consistent with Julia's terminology.

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+    """
+sco(s; process=without_caption_label)
+```
+
+As we can see, the median differs substantially from the mean.
+
+### Mode {#sec:stats_central_mode}
+
+The mean and median can be useful for numerical and ordinal data.
+However, they are ineffective for nominal data, in which our data is comprised of qualitative data (also known as categorical data).
```
rikhuijzer, line 136: Nice fun fact

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+This is where we use the **mode**, defined as **the most frequent value in our data**.
+
+For the mode, we *do not* have a `mode` function inside Julia's standard library `Statistics` module.
+Instead, we need to use the `StatsBase.jl` for less common statistical functions:
```
rikhuijzer, line 140:
```suggestion
Instead, we need to use the `StatsBase.jl` package for less common statistical functions:
```

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+- **Lower row**: **non-normal** distributed data
+
+```jl
+s = """
+    fig = plot_central()
+    caption = "Normal and Non-Normal Distributed Data -- Differences Between Central Tendencies."
+    label = "plot_central"
+    Options(fig; filename=label, caption, label)
+    """
+sco(s)
```
rikhuijzer, line 180:
```suggestion
fig = plot_central()
caption = "Normal and Non-Normal Distributed Data -- Differences Between Central Tendencies."
label = "plot_central"
Options(fig; filename=label, caption, label)
```
This hides the code for the plot from the reader. If you want to show the code for `plot_central`, use `@sco plot_central()` and inside that function:
```
function plot_central()
     fig = ...
     return Options(fig; filename=label, caption, label) # hide
end
```

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+```
+
+You can see that the mean, median and mode are almost the same in the normal distributed data, but they differ a lot in the non-normal distributed data.
```
rikhuijzer, line 183: Super nice example and figure.

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+    df = more_grades()
+    filter!(row -> row.name == "Hank", df)
+    df
+    """
+sco(s; process=without_caption_label)
+```
+
+If we plug Hank's grades into @eq:mad we have to calculate $\operatorname{median}([2, 0, 0])$, so we end up with the middle value in an ordered list which is $0$.
+
+One more, in @fig:plot_dispersion_mad, we have two data distributions:
```
rikhuijzer, line 327: Awesome again

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+Like the mean absolute deviation, **IQR**, since uses the median and percentiles, is also **robust to outliers**.
+
+As usual, in @fig:plot_dispersion_iqr, we have two data distributions:
+
+- **Upper row**: **normal** distributed data
+- **Lower row**: **non-normal** distributed data
+
+```jl
+s = """
+    fig = plot_dispersion_iqr()
```
rikhuijzer, line 378: :heart:

contents/stats.md
```diff
@@ -0,0 +1,445 @@
...
+
+### Variance and Standard Deviation {#sec:stats_dispersion_varstd}
+
+The first dispersion measure that we will cover are variance and standard deviation.
+We will be covering both measures together because they have a profound relationship.
+**The variance is the square of the standard deviation and the standard deviation is the square root of the variance.**
+
+Let's start with the variance. **Variance is mean of the squared difference between measurements and their average value**:
+
+$$ \operatorname{Var}(x) = \frac{1}{n-1} \sum^n_{i=1} (x_i - \bar{x})^2. $$ {#eq:variance}
```
rikhuijzer, line 214: In the equations, `median` was in all lowercase, `Var` and `Cov` start with an uppercase and `MAD` is in all uppercase. Maybe, this can be made more consistent? I think that `Var` and `var` are both used in practice in equations depending on the domain.

src/stats.jl
```diff
@@ -0,0 +1,441 @@
+u = LinRange(0, 2π, 72)
+a, b = 5.0, 2.0
+function ellipse(u; a = 2, b = 3, h = 0, k = 0)
+    Point2f(h + a/2*cos(u), k + b/2*sin(u))
+end
+
+# https://github.com/dronir/Bezier.jl
```
rikhuijzer, line 7: Can the `Bezier.jl` code be put in it's own file? `include("bezier")`

src/stats.jl
```diff
@@ -0,0 +1,441 @@
...
+function statistics_graph()
+    u = LinRange(0, 2π, 72)
+    a, b = 5.0, 2.0
+    # arrow lines 
+    x0 = [1, 0] # start point
+    x1 = [0.75, -5] # end point
+    t0 = [2, 1.0] # starting tangent vector
+    t1 = [-2, 1.0] # end tangent vector
+    curve = BezierCurve(x0, x1, t0, t1)
+    T = range(0, 1 ; length=100)
```
rikhuijzer, line 48:
```suggestion
    T = range(0, 1; length=100)
```

src/stats.jl
```diff
@@ -0,0 +1,441 @@
...
+    ax1 = Axis(fig[1, 1]; limits=((3, 20), nothing))
+    ax2 = Axis(fig[2, 1]; limits=((3, 20), nothing))
+    d1 = Distributions.Normal(10)
+    d2 = LogNormal(log(10), log(1.5))
+    density!(
+        ax1, rand(d1, 1_000); strokewidth=1.5, strokecolor= (:black, 0.5), color=(:silver, 0.15)
+    )
+    density!(
+        ax2, rand(d2, 1_000); strokewidth=1.5, strokecolor= (:black, 0.5), color=(:grey, 0.25)
+    )
```
rikhuijzer, line 94:
```suggestion
strokewidth = 1.5
strokecolor = (:black, 0.5)
dens(ax, d, color) = density!(ax, rand(d, 1_000); strokewidth, strokecolor, color)
dens(ax1, d1, (:silver, 0.15))
dens(ax2, d2, (:grey, 0.25))
```

src/stats.jl
```diff
@@ -0,0 +1,441 @@
...
+        ymax=Distributions.pdf(d1, mean(d1)) / ylim_ax1,
+        color= :dodgerblue,
+        linewidth=3,
+        linestyle = :solid,
+        label=L"\mu",
+    )
+    vlines!(
+        ax1,
+        [mean(d1) - std(d1), mean(d1) + std(d1)];
+        ymax=Distributions.pdf(d1, [mean(d1) - std(d1), mean(d1) + std(d1)]) ./ ylim_ax1,
```
rikhuijzer, line 193:
```suggestion
msd = [mean(d1) - std(d1), mean(d1) + std(d1)]
    vlines!(
        ax1,
        msd;
        ymax=Distributions.pdf(d1, msd) ./ ylim_ax1,
```

src/stats.jl
```diff
@@ -0,0 +1,441 @@
...
+function plot_corr()
+    Random.seed!(123)
+    CairoMakie.activate!() # hide
+    fig = Figure(; resolution=(600, 600))
+    corrs = [0.5, -0.5, 0.8, -0.8]
+    ds = [MvNormal([1 i; i 1]) for i in corrs]
+    d0 = MvNormal(2, 1)
+    ax1 = Axis(
+               fig[1, 1:2]; title="Correlation = $(corrs[1])", titlesize=20, limits=((-2, 2), (-2, 2))
+               )
```
rikhuijzer, line 410: This is something which only an automated formatter can come up with :laughing:

rikhuijzer (commented):
Very nice

src/stats.jl
storopoli: Yes

contents/stats.md
storopoli: Let's wait for @lazarusA opinions on math stuff.

rikhuijzer (approved):

pandoc/references.bib
```diff
@@ -209,3 +209,13 @@ @article{wickham2011split
...
 }
+
+@book{kolmogorovFoundationsTheoryProbability1933,
+  title = {Foundations of the {{Theory}} of {{Probability}}},
+  shorttitle = {Foundations of the {{Theory}} of {{Probability}}},
+  author = {Kolmogorov, Andrey Nikolaevich},
+  date = {1933},
+  publisher = {{Julius Springer}},
+  location = {{Berlin}},
+  file = {/home/storopoli/Zotero/storage/NMSXCGSW/books.html}
```
rikhuijzer, line 220:
```suggestion
```

contents/stats_distributions.md
```diff
@@ -0,0 +1,62 @@
+## Distributions {#sec:stats_dist}
+
+Before defining distributions, we must talk about **probability**.
+
+**The probability of an event is a real number ($\in \mathbb{R}$) between 0 and 1, where, roughly, 0 indicates the impossibility of the event and 1 indicates the certainty of the event**.
```
lazarusA, line 5: R is the set of real numbers, but then what is a set? Probably a reference to R?

contents/stats_distributions.md
```diff
@@ -0,0 +1,62 @@
...
+The greater the likelihood of an event, the more likely it is that the event will occur.
+A simple example is the tossing of a fair (impartial) coin.
+Since the coin is fair, both results ("heads" and "tails") are equally likely; the probability of "heads" is equal to the probability of "tails"; and since no other result is possible, the probability of "heads" or "tails" is $\frac{1}{2}$ (which can also be written as 0.5 or 50%).
+
+Regarding notation, **we define $A$ as an event and $P(A)$ as the probability of event $A$**, thus:
+
+$$ \{P(A) \in \mathbb{R} : 0 \leq P(A) \leq 1 \}. $$ {#eq:probability}
+
+This means the "probability of the event to occur is the set of all real numbers between 0 and 1; including 0 and 1".
```
lazarusA, line 14: is the set of all real numbers in the close interval $[0,1]$. (but probably too formal).

contents/stats_distributions.md
```diff
@@ -0,0 +1,62 @@
...
+In addition, we have three axioms, originated from @kolmogorovFoundationsTheoryProbability1933:
+
+1. **Non-negativity**: For all $A$, $P(A) \geq 0$. Every probability is positive (greater than or equal to zero), regardless of the event.
+2. **Additivity**: For two mutually exclusive $A$ and $B$ (cannot occur at the same time): $P(A) = 1 - P(B)$ and $P(B) = 1 - P(A)$.
```
lazarusA, line 18:
```suggestion
2. **Additivity**: For two mutually exclusive events $A$ and $B$ (cannot occur at the same time): $P(A) = 1 - P(B)$ and $P(B) = 1 - P(A)$.
```

contents/stats_distributions.md
storopoli: This section is not ready. I've just dumped some stuff from https://github.com/storopoli/Bayesian-Julia to polish afterwards.

contents/stats_distributions.md
storopoli: We can stick with this definition because it plays nice with the Kolmogorov axioms below. But, yes, it is the set of all real number between 0 and 1 (both inclusive).

I will add a more intuitive explanation and also some sugarcoating stuff to explain better.

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+* measures of **central tendencies**
+* measures of **dispersion**
+* measures of **dependence**
+* **probability distributions**
+* statistical **visualization**
+
+Statistics is important because it is a **tool to make sense of data**.
+With the abundant availability of data, we are often overwhelmed by numbers.
+Statistics offers a way to comprehend, summarize and infer information from data.
+We believe that every data scientist should have a basic understanding of statistics and how to perform simple statistical operations.
```
lazarusA, line 19:
```suggestion
We believe that every data scientist needs to have a basic understanding of statistics and how to perform simple statistical operations.
```
otherwise I agree with @rikhuijzer 's suggestion.

storopoli: Ok let's go with "needs".

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+
+We can divide statistics into two broad categories: **descriptive** and **inferential**.
+**Descriptive statistics summarizes and quantifies the characteristics of a given observed data**.
+Common metrics are: mean, median, mode, standard deviation, variance, correlation, percentiles.
+
+**Inferential statistics allows generating inferences** (statements) **from observed data about the data generation process**.
```
lazarusA, line 25:
```suggestion
**Inferential statistics allows generating inferences** (statements) **from observed data about the data generation process**.
In @fig:stats_vs_prob we summarize the relationship between data generating process and observed data.
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+All phenomena have an underlying data generating process that describes how the data is being generated.
+For example, in a soccer game, a scored goal can be explained by an underlying process: a tactic, error, struck of luck; or a mix of those.
+If we know a phenomenon's data generating process, we can use probability to simulate possible scenarios given certain aspects.
+Most of the time, especially in applied sciences, we do not have full knowledge of the data generating process.
+Given the observed data, we can retrace our way back to the data generating process.
+This process is known as **statistical inference**.
+Given some data, we can *infer* what are the aspects of the underlying data generating process.
+This is the realm of inferential statistics.
+
+In @fig:stats_vs_prob we summarize the relationship between data generating process and observed data.
```
lazarusA, line 35:
```suggestion
```

contents/stats.md
lazarusA: Yes, a little bit earlier will be better. See suggestions.

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+
+```jl
+fig = statistics_graph()
+caption = "Statistics vs Probability"
+label = "stats_vs_prob"
+Options(fig; filename=label, caption, label)
+```
+
+In this chapter, we will cover only descriptive statistics.
+Inferential statistics is an important and fundamental component of applied sciences, but its scope is too broad. Let's begin on simple ways to summarize our data with central tendencies.
```
lazarusA, line 47:
```suggestion
Inferential statistics is an important and fundamental component of applied sciences, but its scope is too broad. So, let's begin on simple ways to summarize our data by using central tendencies.
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+### Mean {#sec:stats_central_mean}
+
+The most common central tendency measure is the mean.
+**The mean, also known as average, is the sum of all measurements divided by the number of observations** and we generally denote it as "x bar":
+
+$$ \bar{x} = \frac{1}{n} \sum^n_{i=1} x_i = \frac{x_1 + x_2 + \cdots + x_n}{n}, $$ {#eq:mean}
+
+where $\bar{x}$ is the sample mean of the variable $\mathbf{x} = x_1, \cdots, x_n$.
+Often, we see the mean denoted with the **Greek letter $\mu$**, for example $\bar{x} = \mu_x$.
+Additionally, the mean is also known as the **expectation** which is represented by the operator $\operatorname{E}$, thus the mean of a variable $x$ becomes a $\operatorname{E}(x)$[^expectation].
```
lazarusA, line 62:
```suggestion
Additionally, the mean is also known as the **expectation** which is represented by the operator $\operatorname{E}$, thus the mean of a variable $x$ becomes  $\operatorname{E}(x)$[^expectation].
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+So, bear in mind that you might find different notations for the mean.
+
+[^expectation]: technically speaking, the expectation is defined as either a sum for discrete variables or an integral for continuous variables. In contrast with the mean which is the sum of the variables' values divided by number of values, the expectation can be defined over an infinite domain while the mean only works for finite domains.
```
lazarusA, line 65: `mean only works for finite domains`, is this in the reference?, don't see it and not sure is this always true.

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+You might be wondering: "which central tendency shall I use? Mean? Median? Mode?".
+Here is our advice:
+
+- For data that do *not* have outliers, use the **mean**
+- For data that *do* have outliers, use the **median**
+- For categorical/nominal data, use the **mode**
+
+## Dispersion Measures {#sec:stats_dispersion}
+
+In statistics, **dispersion is a measure of how much spread certain observations are from a central tendency**.
```
lazarusA, line 198:
```suggestion
In statistics, **dispersion is a measure of how much spread certain observations have from a central tendency**.
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+
+### Variance and Standard Deviation {#sec:stats_dispersion_varstd}
+
+The first dispersion measure that we will cover are variance and standard deviation.
+We will be covering both measures together because they have a profound relationship.
+**The variance is the square of the standard deviation and the standard deviation is the square root of the variance.**
+
+Let's start with the variance. **Variance is mean of the squared difference between measurements and their average value**:
+
+$$ \operatorname{Var}(x) = \frac{1}{n-1} \sum^n_{i=1} (x_i - \bar{x})^2. $$ {#eq:variance}
```
lazarusA, line 211:
```suggestion
$$ \operatorname{var}(x) = \frac{1}{n-1} \sum^n_{i=1} (x_i - \bar{x})^2. $$ {#eq:variance}
```

contents/stats.md
lazarusA: I say, let's do lower case. Less confusion later when calling the function `cov` or `var`.

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+    """
+sco(s; process=without_caption_label)
+```
+
+We can see that Sally has the highest dispersion in her grades measured by its variance.
+
+**The squared deviation is the square root of the mean of the squared difference between measurements and their average value**.
+Or in more simple words: it is the **square root of the variance**:
+
+$$ \sigma(x) = \sqrt{\frac{1}{n-1} \sum^n_{i=1} (x_i - \bar{x})^2}. $$ {#eq:std}
```
lazarusA, line 242: Ok. See, using $\sigma$ feels more natural, probably is better to do the same for variance $\sigma^2$ in the previous paragraph.

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+Options(fig; filename=label, caption, label)
+```
+
+We can see that the mean $\mu$ is slightly shifted towards to the right by the few influential observations and that the dispersion measured by the $\pm 1\sigma$ (one standard deviation) away from the mean inherits the bias from the mean.
+
+### Mean Absolute Deviation {#sec:stats_dispersion_mad}
+
+Since variance and standard deviation uses the mean in their mathematical formulation, they are also **sensitive to outliers**.
+This is where a **dispersion measure that uses the median instead of the mean** would be useful.
+This is exactly the case of the **median absolute deviation which is defined as the median of the absolute difference between measurements and their median value**.
```
lazarusA, line 282:
```suggestion
This is exactly the case of the **Median Absolute Deviation (MAD) which is defined as the median of the absolute difference between measurements and their median value**.
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+Also known as its acronym MAD, it is an extreme robust dispersion measure since it uses twice the median to calculate first the central tendency followed by the difference between observations and their central distance:
```
lazarusA, line 283:
```suggestion
MAD is an extreme robust dispersion measure since it uses twice the median to calculate first the central tendency followed by the difference between observations and their central distance:
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+# mad
+s = """
+    gdf = groupby(more_grades(), :name)
+    combine(gdf, :grade => mad)
+    """
+sco(s; process=without_caption_label)
+```
+
+We can see that still Sally has the highest grades dispersion, but now Bob's and Alice's dispersion are the same.
+Also note that, by using $\operatorname{MAD}$, Hank's dispersion is zero.
```
lazarusA, line 307:
```suggestion
Also note that, by using $\operatorname{mad}$, Hank's dispersion is zero.
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+    df = more_grades()
+    filter!(row -> row.name == "Hank", df)
+    df
+    """
+sco(s; process=without_caption_label)
+```
+
+If we plug Hank's grades into @eq:mad we have to calculate $\operatorname{median}([2, 0, 0])$, so we end up with the middle value in an ordered list which is $0$.
+
+One more, in @fig:plot_dispersion_mad, we have two data distributions:
```
lazarusA, line 321:
```suggestion
Once more, in @fig:plot_dispersion_mad, we have two data distributions:
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+- **Lower row**: **non-normal** distributed data
+
+```jl
+fig = plot_dispersion_mad()
+caption = "Normal and Non-Normal Distributed Data -- Differences Between Mean Absolute Deviations."
+label = "plot_dispersion_mad"
+Options(fig; filename=label, caption, label)
+```
+
+Note that the MAD is extremely robust against influential observations and, contrary to the standard deviation, does not inherit any bias from the underlying central tendency measure (median).
```
lazarusA, line 333:
```suggestion
Note that the `mad` is extremely robust against influential observations and, contrary to the standard deviation, does not inherit any bias from the underlying central tendency measure (median).
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+MAD can be an effective dispersion measure to non-normal data which few influential observations shift a non-robust central tendency (such as the mean).
```
lazarusA, line 334:
```suggestion
`mad` can be an effective dispersion measure to non-normal data which few influential observations shift a non-robust central tendency (such as the mean).
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+```
+
+Here we can see that the median is *not* influenced by the few influential observations and that the IQR measured by the 1st and 3rd quantiles represents the 50% most probable values of our observations.
+
+### Advice on Dispersion Measures {#sec:stats_dispersion_advice}
+
+You might be wondering: "which dispersion measure shall I use? Variance? Standard Deviation? Mean Absolute Deviation? IQR?".
+Like before, we provide the following advice:
+
+- _Don't_ use **variance**, since it is not an intuitive measure
```
lazarusA, line 381: This is a really strong statement. Probably, not the right approach. Their use might be problem dependent.

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+- For data that do *not* have outliers, use the **standard deviation**
+- For data that *do* have outliers, use either the **mean absolute deviation** or **IQR**
+- For categorical/nominal data, use some sort of **frequency counter**
```
lazarusA, line 384: Need to explain frequency counter? or a footnote?

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+ 
+## Dependence Measures {#sec:stats_dependence}
+
+Now that we covered univariate measures of central tendency and dispersion, we need to talk about bivariate measures.
+**Univariate measures are measures with respect to a single variable**.
+**Bivariate measures are measures with respect to the _relationship_ between two variables**.
+To quantity such relationships we use **dependence measures, which measure somehow the degree of how one variable value depends (or is related) to another variable value**.
```
lazarusA, line 391:
```suggestion
To quantify such relationships we use **dependence measures, which measure the degree of how one variable value depends (or is related) to another variable value**.
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+
+There is a distinction between a dependence *measure* and a dependence *relationship*.
+The first is simply a bivariate measure of how one variable is related to the other.
+On the other hand, a dependence relationship implies a profounder relationship, where not only they are related but information regarding one variable gives us information regarding the other.
```
lazarusA, line 395:
```suggestion
On the other hand, a dependence relationship implies a deeper connection, where not only they are related but information regarding one variable gives us information about the other one.
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+
+There two main measures of dependence and they are practically the same since one is the other constrained to the unit range between -1 and 1.
```
lazarusA, line 397:
```suggestion
There are two main measures of dependence and they are practically the same since one is the other constrained to the unit range between -1 and 1.
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+**Covariance is a measure of the joint variability of two variables**.
+It is the expectation of one variable minus its expectation times another variables minus its expectation:
+
+$$ \operatorname{Cov}(x, y) = \operatorname{E} \left[ \left( x - \operatorname{E}(x) \right) \cdot \left( y - \operatorname{E}(y) \right) \right]. $$ {#eq:covariance}
+
+High positive or negative covariance indicates either a strong positive or negative joint variability, respectively.
+It measures how much one unit increase/decrease in one variable is related to one unit increase/decrease in another variable.
+
+The interpretation of a given covariance demands knowledge about units of measurement from both the underlying variables.
```
lazarusA, line 406:
```suggestion
The interpretation of a given covariance demands knowledge about units of measurement from both underlying variables.
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+So, if we want to analyze covariance between two variables we must have some understanding of how the variables units are measured.
+That is why most of the time we use the **correlation, which is _normalized_ covariance**.
+The correlation is _dimensionless_ since it is constrained to the unit range between -1 and 1.
+Mathematically speaking, the correlation is the covariance divided by the product of the variables' standard deviation and is denoted by the Greek letter $\rho$:
+
+$$ \rho(x, y) = \frac{\operatorname{Cov}(x, y)}{\sigma_x \cdot \sigma_y}, $$ {#eq:correlation}
+
+where $\sigma_x$ and $\sigma_y$ are the standard devations of $x$ and $y$, respectively.
```
lazarusA, line 414:
```suggestion
where $\sigma_x$ and $\sigma_y$ are the standard deviations of $x$ and $y$, respectively.
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+High positive or negative covariance indicates either a strong positive or negative joint variability, respectively.
+It measures how much one unit increase/decrease in one variable is related to one unit increase/decrease in another variable.
+
+The interpretation of a given covariance demands knowledge about units of measurement from both the underlying variables.
+So, if we want to analyze covariance between two variables we must have some understanding of how the variables units are measured.
+That is why most of the time we use the **correlation, which is _normalized_ covariance**.
+The correlation is _dimensionless_ since it is constrained to the unit range between -1 and 1.
+Mathematically speaking, the correlation is the covariance divided by the product of the variables' standard deviation and is denoted by the Greek letter $\rho$:
+
+$$ \rho(x, y) = \frac{\operatorname{Cov}(x, y)}{\sigma_x \cdot \sigma_y}, $$ {#eq:correlation}
```
lazarusA, line 412:
```suggestion
$$ \rho(x, y) = \frac{\operatorname{cov}(x, y)}{\sigma_x \cdot \sigma_y}, $$ {#eq:correlation}
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+
+There is a distinction between a dependence *measure* and a dependence *relationship*.
+The first is simply a bivariate measure of how one variable is related to the other.
+On the other hand, a dependence relationship implies a profounder relationship, where not only they are related but information regarding one variable gives us information regarding the other.
+
+There two main measures of dependence and they are practically the same since one is the other constrained to the unit range between -1 and 1.
+**Covariance is a measure of the joint variability of two variables**.
+It is the expectation of one variable minus its expectation times another variables minus its expectation:
+
+$$ \operatorname{Cov}(x, y) = \operatorname{E} \left[ \left( x - \operatorname{E}(x) \right) \cdot \left( y - \operatorname{E}(y) \right) \right]. $$ {#eq:covariance}
```
lazarusA, line 401:
```suggestion
$$ \operatorname{cov}(x, y) = \operatorname{E} \left[ \left( x - \operatorname{E}(x) \right) \cdot \left( y - \operatorname{E}(y) \right) \right]. $$ {#eq:covariance}
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+In @fig:plot_corr, we can see some correlations and their underlying scatter plot for 50 random generated observations.
+
+```jl
+fig = plot_corr()
+caption = "Correlation for 50 Random Bivariate Observations."
+label = "plot_corr"
+Options(fig; filename=label, caption, label)
+```
+
+As we can see, correlation depicts the **linear association** between variables.
```
lazarusA, line 425:
```suggestion
As we can see, the correlation depicts the **linear association** between variables.
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+The slope of the dashed line shows the correlation between variables.
+The more far from zero and closer to $\pm$ 1 the stronger is the association between variables.
+Finally, the sign of the correlation denotes the type of relationship between variables.
+**Positive correlations** implies **positive relationship**, increase in one variable results in an _increase_ in the other.
```
lazarusA, line 429:
```suggestion
**Positive correlations** implies **positive relationship**, increase in one variable results in an _increase_ on the other one.
```

contents/stats.md
```diff
@@ -0,0 +1,430 @@
...
+**Negative correlations** implies **negative/inverse relationship**, increase in one variable results in an _decrease_ in the other.
```
lazarusA, line 430:
```suggestion
**Negative correlations** implies **negative/inverse relationship**, increase in one variable results in a _decrease_ on the other one.
```

lazarusA (commented):
Ok... done, probably a second round will be good after the whole chapter is finished.

contents/stats.md
storopoli: Do you have a better explanation. I've used this from my head from our conversation. We can use a better term here.

contents/stats.md
storopoli: Ok

contents/stats.md
storopoli: You are right, I will change to something better.

contents/stats.md
storopoli: I don't think so... It's pretty much self-explanatory.

contents/stats.md
storopoli: Changed to avoid

contents/stats.md
lazarusA, line 65:
```suggestion
[^expectation]: technically speaking, the expectation is defined as either a sum for discrete variables or an integral for continuous variables weighted by a probability distribution. In contrast with the mean which is simply the arithmetic sum of the variables' values divided by number of values.
```
I would leave it like this. Regarding domains and existence, finiteness,  of the mean depends on the probability distribution, so better not go into much details here.

storopoli: Thanks! Works great!

lazarusA (approved):

storopoli:
also https://github.com/rikhuijzer/Books.jl/issues/187 needs to be fixed for inline calculation of PDF of a continuous variable

storopoli:
Ok guys, there is only one more section left. But I will include it in a next PR. For this PR I think we are done.
I've finalized the statistics distributions section.

contents/stats_distributions.md
```diff
@@ -0,0 +1,159 @@
...
+$$ X \sim \operatorname{dist}(\theta_1, \theta_2, \dots), $$ {#eq:distribution_notation}
+
+where $\operatorname{dist}$ is the name of the distribution and $\theta_1, \theta_2, \dots$ are the parameters that control how the distribution behaves.
+Every distribution can be "parameterized" by specifying parameters that allow us to tailor some aspects of the distribution for some specific purpose.
+
+Also, notice that we are using upper case $X$ instead of lower case $x$.
+This is a wide used convention in probability and statistics, that **upper case characters denote random variables, and lower case characters denote numerical values of a random variable**.
+More broadly, $X$ is a theoretical distribution, whereas $x$ are the observed values from $X$.
+
+There are several distributions defined in textbooks[^stats_book] and scientific articles[^stats_articles].
```
rikhuijzer, line 39:
```suggestion
There are several distributions defined in textbooks [for example, @bertsekas2000introduction] and scientific articles [for example, @lewandowskiGeneratingRandomCorrelation2009].
```

contents/stats_distributions.md
```diff
@@ -0,0 +1,159 @@
...
+[^stats_book]: we recommend @bertsekas2000introduction.
+[^stats_articles]: one example is the $\operatorname{LKJ}$ distribution [@lewandowskiGeneratingRandomCorrelation2009].
+
+```julia
+using Distributions
+```
+
+### Normal and Non-Normal Distributions {#sec:stats_dist_normal}
+
+In applied sciences, there is a widely used distribution called the **normal distribution**.
```
rikhuijzer, line 52:
```suggestion
The most commonly arising distribution is the **normal distribution**.
It occurs a lot in nature; for example, the height and weight of humans and the size of snowflakes.
```

contents/stats_distributions.md
```diff
@@ -0,0 +1,159 @@
...
+It has the ubiquitous shape of a bell curve and is also known as the *Gaussian* distribution in honor of the mathematician Carl Friedrich Gauss.
+
+This distribution is generally used in the social and natural sciences to represent variables in which their distributions are not known.
+Some phenomena follow a normal distribution, such as human height, human shoe size or test scores.
+
+We urge the reader to perceive that *not all phenomena are normally distributed*.
```
rikhuijzer, line 58:
```suggestion
However, beware that *not all phenomena are normally distributed*.
```

contents/stats_distributions.md
```diff
@@ -0,0 +1,159 @@
...
+For example, income, financial returns, city size, social media followers are examples of important phenomena that, generally, do not follow a normal distribution.
+Instead, they follow what we will call as non-normal distribution, or, as some also denote, a long-tailed distribution.
```
rikhuijzer, line 60:
```suggestion
Instead, they follow what we will call a non-normal distribution, or, as some also denote, a long-tailed distribution.
```

contents/stats_distributions.md
```diff
@@ -0,0 +1,159 @@
...
+
+In @fig:plot_normal_lognormal, we can see the comparison between a distribution that follows a normal distribution and a distribution that is long-tailed, i.e. does *not* follow a normal distribution.
+
+```jl
+fig = plot_normal_lognormal()
+caption = "Normal and Non-Normal Distributed Distributions."
+label = "plot_normal_lognormal"
+Options(fig; filename=label, caption, label)
```
rikhuijzer, line 68: FYI, since yesterday, you can also show the code here without using `# hide` thanks to https://github.com/rikhuijzer/Books.jl/pull/209.

Then, it would be
```
let
    caption = "Normal and Non-Normal Distributed Distributions."
    label = "plot_normal_lognormal"
    pre(fig) = Options(fig; filename=label, caption, label)
    @sco pre=pre plot_normal_lognormal()
end
```
cc @lazarusA

contents/stats_distributions.md
```diff
@@ -0,0 +1,159 @@
...
+```
+
+### Discrete and Continuous {#sec:stats_dist_discrete_continuous}
+
+Regarding distributions we have mainly two types of distributions: **discrete** and **continuous**.
```
rikhuijzer, line 73:
```suggestion
Regarding distributions, we have mainly two types of distributions: **discrete** and **continuous**.
```

rikhuijzer (approved):
It's a bit difficult to see what I have reviewed now and what not, but I think that I've now seen everything.

Lets merge after Lazaro's review :tada: Great work

contents/stats_distributions.md
rikhuijzer: I think I got the syntax right here, but I didn't check

rikhuijzer: I checked. It should be correct.

contents/stats_distributions.md
```diff
@@ -0,0 +1,159 @@
...
+
+This means the "probability of the event to occur is the set of all real numbers between 0 and 1; including 0 and 1".
+In addition, we have three axioms, originated from @kolmogorovFoundationsTheoryProbability1933:
+
+1. **Non-negativity**: For all $A$, $P(A) \geq 0$. Every probability is positive (greater than or equal to zero), regardless of the event.
+2. **Additivity**: For two mutually exclusive events $A$ and $B$ (cannot occur at the same time): $P(A) = 1 - P(B)$ and $P(B) = 1 - P(A)$.
+3. **Normalization**: The probability of all possible events $A_1, A_2, \dots$ must add up to 1: $\sum_{n \in \mathbb{N}} P(A_n) = 1$; or, in the case of a continuous variable, integrate to 1: $\int^\infty_{-\infty} f(x) dx = 1$.
+
+Now we are ready to talk about distributions.
+A **distribution is a function which maps real numbers to a probability**:
```
lazarusA, line 22: Simply put: a distribution is a just a collection of scores (counts) in a variable(observations). Later,  a **probability distribution** is defined by a function that will map each observation to a given probability, or course after proper normalisation of the distribution.

contents/stats_distributions.md
```diff
@@ -0,0 +1,159 @@
...
+Analogously, in the continuous case, the inputs have a "density", thus we characterize continuous distributions with the **probability density function** (pdf).
+
+$$ \operatorname{pdf} = P(a \leq X \leq b) = \int_a^b f(x) dx $$ {#eq:pdf}
+
+Since a continuous distribution takes uncountably infinite this means that the pdf of a specific value is always *zero* and the pdf is only defined on an interval.
+That's why we have an integral in @eq:pdf.
+
+For example, in @fig:plot_pdf, we have a normal distribution with mean 0 and standard deviation 1.
+Notice that, since this is a valid probability, the gray shaded area that represents all the possible values integrates to 1.
+The red shaded area is the probability, under this distribution, of we observing values from 1 to 2 and we need to calculate the total area.
```
lazarusA, line 114:
```suggestion
The red shaded area is the probability, under this distribution, of observing values from 1 to 2 and we need to calculate the total area.
```

lazarusA (approved):

contents/stats_distributions.md
storopoli: What's the difference. The other footnotes are like the way I wrote. I think we should be consistent

contents/stats_distributions.md
storopoli: Thanks done in 03a33c71f3a147f7b31cc9cf8064fc867f6c30f1

storopoli:
Thanks guys I will be merging this and will work on the last section of the chapter

PR #121, opened 2021-09-19
Document using versus import

rikhuijzer:
@lazarusA, in response to using Plots and Makie at the same time. I think that the most future proof solution is to use

```
@reexport begin

using Plots:
    plot,
    [...]
using Makie:
    density,
    [...]

end # @reexport
```

The rationale is added to the notation section and can also be read in the [JuMP Style Guide](https://jump.dev/JuMP.jl/stable/developers/style/#using-vs.-import).

As a proof-of-concept, I've implemented it for `DataFrames.jl` and `XLSX.jl` already.

This pattern is also used at other places. For example, in [`JETTest.jl`](https://github.com/aviatesk/JETTest.jl/blob/master/src/JETTest.jl).

### DIFF
--- a/pandoc/references.bib
+++ b/pandoc/references.bib
@@ -70,7 +70,7 @@
 @misc{jump2021using,
   title = {JuMP Style Guide},
   year = 2021,
-  howpublished = {\url{https://jump.dev/JuMP.jl/v0.21/developers/style/#using-vs.-import}},
+  url = {https://jump.dev/JuMP.jl/v0.21/developers/style/#using-vs.-import},
   note = {Accessed: 2021-09-19}
 }
 
--- a/pandoc/references.bib
+++ b/pandoc/references.bib
@@ -67,6 +67,13 @@
   howpublished = {Zenodo}
 }
 
+@misc{jump2021using,
+  title = {JuMP Style Guide},
+  year = 2021,
+  howpublished = {\url{https://jump.dev/JuMP.jl/stable/developers/style/#using-vs.-import}},
+  note = {Accessed: 2021-09-19}
+}
+
 @article{khan2014big,
   title     = {Big data: survey, technologies, opportunities, and challenges},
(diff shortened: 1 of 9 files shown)


storopoli:
We might need to do this for all sections.
Also, why is the build failing?

src/JDS.jl
```diff
@@ -4,28 +4,52 @@ import XLSX
...
+using FileIO
+using GLMakie
+using GeometryBasics
+using InteractiveUtils
+using LaTeXStrings
+using LinearAlgebra
+using Makie
+using Random
+using Statistics
+using StatsBase
```
storopoli, line 47:
```suggestion
using StatsBase:
    mad,
    mode
```

storopoli (approved):

pandoc/references.bib
```diff
@@ -70,7 +70,7 @@ @misc{johnmyleswhiteJuliaDataDataFramesJl2020
 @misc{jump2021using,
   title = {JuMP Style Guide},
   year = 2021,
-  howpublished = {\url{https://jump.dev/JuMP.jl/stable/developers/style/#using-vs.-import}},
```
storopoli: Hardcoding versions?

rikhuijzer: I just did that in 7acc2f9

rikhuijzer: For the docs it makes sense, right?

storopoli: What if in year from now JUMP is on v1? Do we need to keep track of JuMP versions?

rikhuijzer: It's only a ref to the information stored in that style guide. That information is still valid when JuMP is at a newer version.

storopoli: OK

storopoli:
“Prefer to load symbols” sounds like `:A`. I think we need to explain what “symbols” are in this context.

storopoli (approved):

lazarusA (approved):

PR #123, opened 2021-09-19
Add front cover to homepage

rikhuijzer:
@lazarusA, I've moved your code into the package and made some minor changes to get it working.


rikhuijzer:
> Nice `devcontainers` are awesome

I've added it because `GLMakie` doesn't work on NixOS :\

lazarusA:
Ok. I really don't like this one 🥲(it does not look good enough), so I would work a little bit more in the front cover, and update that image ASAP.

rikhuijzer:
> Ok. I really don't like this one smiling_face_with_tear(it does not look good enough), so I would work a little bit more in the front cover, and update that image ASAP.

Okay. I've now hidden the image from the front page. Is it then okay to merge?

lazarusA (approved):
Nice, let's have in the main branch.

PR #124, opened 2021-09-20
Format trailing whitespace

rikhuijzer:
I've tried to do it via JuliaFormatter.jl. Unfortunately, that one is too quite aggressive. For now, I'll wait on https://github.com/domluna/JuliaFormatter.jl/pull/470.

The plan is to add a CI check for trailing whitespaces and other minor things.


storopoli:
Great I also subscribed to the https://github.com/domluna/JuliaFormatter.jl/pull/470 which I will use extensively in the future.

domluna:
@rikhuijzer what part was too aggresive?

rikhuijzer:
> @rikhuijzer what part was too aggresive?

Hey, @domluna, thank your for your interest. I couldn't figure out how to set JuliaFormatter.jl to only fix trailing whitespace and, let's say, spaces in keyword arguments. Whatever I tried, the formatter was changing all kinds of things like putting things on newlines etc. Do you want me to open an issue at your repo?

domluna:
oh i see, the PR you linked will fix most of that but there will still be cases that will be reformatted, for example

```
function foo()
   body   end
```

->

```
function foo()
    body
end
```

rikhuijzer:
> oh i see, the PR you linked will fix most of that but there will still be cases that will be reformatted, for example
> 
> ```
> function foo()
>    body   end
> ```
> 
> ->
> 
> ```
> function foo()
>     body
> end
> ```

That sounds very reasonable :+1:

PR #126, opened 2021-09-21
[Statistics] Final Section on Statistical Visualizations

storopoli:
- [x] Statistical Visualizations
	- [x] Box plots
	- [x] Histograms
	- [x] Density plots
	- [x] Anscombe Quartet

Closes #112 

OBS: I had to name a section on Makie chapter because I need to reference it in this chapter (cc @lazarusA)

PR #128, opened 2021-09-22
Fix trailing whitespace

rikhuijzer:
Add a simple trailing whitespace check.

For some reason, CI in #124 is broken. So, that's why this is a new PR for the same thing.

To locally remove all whitespace, use:

```
$ julia --project=format

julia> using FormatJDS

julia> remove_trailing_whitespace()

```

EDIT: I've tried only starting `CI` if `Format` succeeds to save some build minutes. Unfortunately, that should be done via `workflow_run` and that appears to be a bit weird. There even exist Actions to workaround missing functionality (for example, https://github.com/ahmadnassri/action-workflow-run-wait).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,13 +1,11 @@
 name: CI
 
 on:
-  workflow_run:
+  push:
     branches:
       - main
-    workflows:
-      - Format
-    types:
-      - completed
+  pull_request:
+  workflow_dispatch:
 
 jobs:
   build-and-deploy:
@@ -16,9 +14,6 @@ jobs:
       - uses: actions/checkout@v2
         with:
           persist-credentials: false
-
-      - name: Only continue if workflow_run workflows were successful.
-        uses: ahmadnassri/action-workflow-run-wait@v1
 
       - uses: julia-actions/setup-julia@v1
         with:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,15 +1,15 @@
 name: CI
 
 on:
-  push:
-    branches:
-      - main
-  pull_request:
-  workflow_dispatch:
+  on:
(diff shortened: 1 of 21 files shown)


rikhuijzer:
The whitespace check works. In 0976f04, the files contain a whitespace and the Julia process exits with "Error: Process completed with exit code 1".

.github/workflows/CI.yml
```diff
@@ -8,8 +8,8 @@ on:
   workflow_dispatch:
 
 jobs:
-  build-and-deploy:
-    runs-on: ubuntu-20.04
+  BuildAndDeploy:
+    runs-on: ubuntu-latest
```
storopoli, line 12: Why `ubuntu-latest`?

src/bezier.jl
```diff
@@ -5,7 +5,7 @@ function ellipse(u; a = 2, b = 3, h = 0, k = 0)
 end
 
 # https://github.com/dronir/Bezier.jl
-# it's not longer in the registery 
+# it's not longer in the registery
```
storopoli, line 8:
```suggestion
# it's not longer in the registry
```

storopoli (approved):
Great! We might get a nasty merge conflict when we finish #126 but it is for a good cause.

.github/workflows/CI.yml
rikhuijzer: No reason in particular. Just so that it's more future proof. In my experience, both work flawless.

storopoli: I find the opposite 😂 when you use a OS that you don't know how it will perform in the future is less future proof. Using ubuntu-20.04 will break but you can have more control and no surprises.

rikhuijzer: You're right. I've set both jobs to ubuntu-20.04.

PR #129, opened 2021-09-22
Show cover on front page

rikhuijzer:
As mentioned in b5dcea4befb5a2910f8b2eb8772ca0ef629aeb90, 506e24078ea60e95b13a030ee6f5feabbba05523 and #123. This PR shows the cover on the front page. The image isn't centered unfortunately. It's a bug in Books.jl: https://github.com/rikhuijzer/Books.jl/issues/217. I'll fix that.

This is the full size image. To avoid the 1.4 MB image to slow down the homepage loading, I've set `loading="lazy"` (more info at https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading).

<br>
<br>

image

### DIFF
--- a/contents/index.md
+++ b/contents/index.md
@@ -36,7 +36,7 @@ Of course, data science is about more things than just tables, basic statistics 
 We want to cover more topics, but we have scheduled them for the second edition of the book.
 For now, the planned topics for the second edition are:
 
-- More statistics (probably, descriptive statistics, `Distributions.jl`)
+- More statistics
 - Plotting via `AlgebraOfGraphics.jl`.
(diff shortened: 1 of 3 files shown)


storopoli:
I think that the title should be in a bigger font size.
Do we need a subtitle?

Ccing @lazarusA

lazarusA:
@storopoli add that to tomorrow's agenda. 
- title size, probably another 10 to the current textsize would be enough. 
- Subtitle or not? 

@rikhuijzer using lazy loading is a good choice... almost always.

lazarusA (approved):

PR #132, opened 2021-09-24
Book Cover

storopoli:
I've cleaned it up.

What needs to be done:
1. The two pipes from 2nd and 3rd row need to have a little padding to the left (maybe 3-4 spaces)

I also renamed the axis to something like ax12 for `Axis(fig[2,3])` which for me is better than the serial ax1 to ax19 or something like that.

I also used `GLMakie`, the original code was using `CairoMakie`. Maybe we need to change that back.

Here is the current book cover from this PR

front_cover2

### DIFF
--- a/src/front-cover.jl
+++ b/src/front-cover.jl
@@ -40,7 +40,7 @@ colorSides =  vec(valsSides[end,:])
 Return the Julia Data Science book front cover.
 """
 function front_cover()
-    GLMakie.activate!()
+    CairoMakie.activate!()
     with_theme(theme_black()) do
         # Figure
         fig = Figure(resolution=(1768,2652))
--- a/src/front-cover.jl
+++ b/src/front-cover.jl
@@ -1,4 +1,5 @@
 const NOTO_SANS_BOLD = assetpath("fonts", "NotoSans-Bold.ttf")
+const JuliaColors = Colors.JULIA_LOGO_COLORS
 Random.seed!(123)
 
 myrand(x,y,z) = rand()
@@ -37,7 +38,6 @@ colorSides =  vec(valsSides[end,:])
 
 Return the Julia Data Science book front cover.
 """
-
 function front_cover()
     GLMakie.activate!()
     with_theme(theme_black()) do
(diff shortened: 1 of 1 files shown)


src/front-cover.jl
```diff
@@ -38,208 +38,139 @@ colorSides =  vec(valsSides[end,:])
 Return the Julia Data Science book front cover.
 """
 function front_cover()
-    CairoMakie.activate!()
+    GLMakie.activate!()
```
storopoli, line 41: Do we need to change back to CairoMakie? @lazarusA

rikhuijzer:
Yes. I like the simplification.

Can I request a few more changes or do you want to do that in a new PR?

My suggestions:

- Author names (centered) below each other. like Modern Data Science with R, other chapman & Hall books and Introduction to Algorithms by Cormen et al.
- Author names in one of the Julia colors? Blue is very nice with black, see Deep Learning by Kelleher. This book also has a nice suggestion for the placement of the title and author name
- The issue about the size isn't yet solved, I think. Pipes are too small still

storopoli:
> Can I request a few more changes or do you want to do that in a new PR? 

Yes, this is the place
 
> * Author names (centered) below each other. like Modern Data Science with R, other chapman & Hall books and Introduction to Algorithms by Cormen et al.

Ok.

> * Author names in one of the Julia colors? Blue is very nice with black, see Deep Learning by Kelleher. This book also has a nice suggestion for the placement of the title and author name

I need @lazarusA's help. Can we do something like:
```julia
Label(fig[3, 3:end], "Jose Storopoli", color = JuliaPurple,
            textsize = 60, tellheight = false)
Label(fig[3, 3:end], "\nRik Huijzer", color = JuliaRed,
            textsize = 60, tellheight = false)
Label(fig[3, 3:end], "\nLazaro Alonso", color = JuliaGreen,
            textsize = 60, tellheight = false)
```
> * The issue about the size isn't yet solved, I think. Pipes are too small still
The 1st column pipes are size 36 and the other are 28 we can up those a little bit.

rikhuijzer (approved):

src/front-cover.jl
lazarusA: I would say yes.

src/front-cover.jl
```diff
@@ -38,208 +40,146 @@ colorSides =  vec(valsSides[end,:])
...
```
lazarusA, line 43:
```suggestion
    CairoMakie.activate!()
```

lazarusA (approved):

PR #134, opened 2021-09-25
Fix CI

rikhuijzer:
The switch to the Tufte template is very breaking, so let's stop running `Books.jl#main`. This should fix CI again.


storopoli (approved):
Also this is aligned with Books.jl v1

rikhuijzer:
Thanks for the review. I've changed my mind :sweat_smile: Fixing the new template is pretty okay. I'm almost done.

PR #136, opened 2021-09-27
Change wording to be less strong

rikhuijzer:
I would say that it's up to the reader to decide when to use the pipe and when not. There are cases where the pipe indeed makes more sense for `CSV.File`, but there are also cases when not. Therefore, calling it "more preferred idiomatic" is a bit too strong.


storopoli (approved):
less is more, agreeded!

PR #139, opened 2021-09-27
change name datavis to Data Visualization, plus some code clean up.

lazarusA:
Please take a look. I changed the figure sizes for most plots in order to have a better looking pdf version. 
@storopoli I also touch a little bit statistics, the diagram, take a look, I think it looks better in that size, and the same should happened for the other figures. 
@rikhuijzer I was not able to apply a different figures size for the themes function. `with_theme(demo_themes, theme_dark())`, https://juliadatascience.io/themes , the Options function could not take the width argument. 

Additionally, I did some clean up of the code, now is shorter, but also uses almost all 92 characters per line, otherwise there is a lot of white space in the pdf version.  Eventually this should improve.

### DIFF
--- a/contents/data_vis_makie_cairo.md
+++ b/contents/data_vis_makie_cairo.md
@@ -10,11 +10,12 @@ CairoMakie.activate!()
 ```jl
 s = """
     CairoMakie.activate!() # hide
-    scatterlines(1:10, 1:10)
+    fig, = scatterlines(1:10, 1:10)
+    #save("firstplot.png", fig) 
     label = "firstplot" # hide
     caption = "First plot." # hide
-    link_attributes = JDS.SMALL_IM_ATTR # hide
-    Options(current_figure(); filename=label, caption, label, link_attributes) # hide
+    link_attributes = "width=60%" # hide
+    Options(fig; filename=label, caption, label, link_attributes) # hide
     """
 sco(s)
 ```
--- a/contents/data_vis_makie_themes.md
+++ b/contents/data_vis_makie_themes.md
@@ -26,6 +26,7 @@ s = """
         with_theme(demo_themes, theme_minimal())
         with_theme(demo_themes, theme_light())
     ] # hide
+   #link_attributes = "width=60%" # hide
     Options.(objects, filenames) # hide
     """
 sco(s)
@@ -48,7 +49,8 @@ s = """
     with_theme(plot_with_legend_and_colorbar, publication_theme())
     label = "plot_with_legend_and_colorbar" # hide
     caption = "Themed plot with Legend and Colorbar." # hide
-    Options(current_figure(); filename=label, label, caption) # hide
+    link_attributes = "width=60%" # hide
+    Options(current_figure(); filename=label, label, caption, link_attributes) # hide
     """
 sco(s)
 ```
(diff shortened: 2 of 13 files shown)


contents/data_vis_makie_cairo.md
```diff
@@ -10,11 +10,12 @@ CairoMakie.activate!()
 ```jl
 s = """
     CairoMakie.activate!() # hide
-    scatterlines(1:10, 1:10)
+    fig, = scatterlines(1:10, 1:10)
+    #save("firstplot.png", fig)
```
rikhuijzer, line 14:
```suggestion
```

This shouldn't be here, I guess.

contents/data_vis_makie_themes.md
```diff
@@ -26,6 +26,7 @@ s = """
         with_theme(demo_themes, theme_minimal())
         with_theme(demo_themes, theme_light())
     ] # hide
+   #link_attributes = "width=60%" # hide
```
rikhuijzer, line 29: Ah, yes I haven't implemented broadcasting `Options` for anything except filename.

You can manually create an `Vector{Options}` with something like
```
[Options(fig; link_attributes) for fig in objects]
```

contents/julia_basics.md
```diff
@@ -451,7 +451,7 @@ scob(s)
 A lot of times, we don't care about the name of the function and want to quickly make one.
 What we need are **anonymous functions**.
 They are used a lot in Julia's data science workflow.
-For example, when using `DataFrames.jl` (@sec:dataframes) or `Plots.jl` (@sec:datavisPlots), sometimes we need a temporary function to filter data or format plot labels.
+For example, when using `DataFrames.jl` (@sec:dataframes) or `Plots.jl` (@sec:DataVisualizationPlots), sometimes we need a temporary function to filter data or format plot labels.
```
rikhuijzer, line 454:
```suggestion
For example, when using `DataFrames.jl` (@sec:dataframes) or `Plots.jl` (@sec:plots), sometimes we need a temporary function to filter data or format plot labels.
```
The section label "plots" would have been enough, I think. The whole point of a section label is that it may differ from the section name (and can be more stable).

But, this comment is nitpicking. It doesn't really matter.

contents/stats.md
```diff
@@ -40,7 +40,8 @@ And by using the observed data we can use inference to gain knowledge about the
 fig = statistics_graph()
 caption = "Statistics vs Probability"
 label = "stats_vs_prob"
-Options(fig; filename=label, caption, label)
+link_attributes = "width=60%" # hide
```
rikhuijzer, line 43:
```suggestion
link_attributes = "width=60%"
```
It is not necessary here, right?

rikhuijzer (commented):
Comments for everything except `src/makie.jl`. I need more time to look in depth into that.

rikhuijzer:
Hmm the whitespace error message isn't particularly helpful. To quickly fix it you can do the following:
```
$ julia --project=format

julia> using FormatJDS

julia> remove_trailing_whitespace()

```

lazarusA:
> Comments for everything except `src/makie.jl`. I need more time to look in depth into that.

ahh yes, that file shouldn't be there... don't look at that one.

contents/julia_basics.md
storopoli: If we accept @rikhuijzer's suggestion, we need to also change in the other files such as `preface.md'

contents/stats.md
storopoli: Yeah, it will not output anything since it isn't inside a string and we are not calling `sco(s)`.

contents/julia_basics.md
lazarusA: Indeed, for now I did changed it everywhere, so it should work. If @rikhuijzer commits the suggestions in all other places I'm fine with it. :D

contents/data_vis_makie_themes.md
lazarusA: I tried without success :( . I will tried another day.

storopoli (approved):
Ok the change in the `stats.md` figure is fine. Thanks!

PR #142, opened 2021-09-29
Fix pipe alignment in front cover

lazarusA:
this should fix some alignment issues regarding the pipes. Additionally, I did a print for the previous one and the lack of grid lines makes the whole cover a little bit dull.

### DIFF
--- a/src/front-cover.jl
+++ b/src/front-cover.jl
@@ -40,8 +40,10 @@ colorSides =  vec(valsSides[end,:])
 Return the Julia Data Science book front cover.
 """
 function front_cover()
-    CairoMakie.activate!()
-    with_theme(theme_black()) do
+    CairoMakie.activate!() # probably it will be good to have to version, one black and one white
+    with_theme(theme_black(); Axis = (; ygridcolor = :grey90, xgridcolor = :grey90,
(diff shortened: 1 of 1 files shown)


lazarusA:
now it looks like this: 

test_front

rikhuijzer (approved):

storopoli:
Why the grids? I've removed them since they do add unwanted complexity for the cover and steers the eyes away from the images.

Also the pipe from the 1st to 2nd row should be more to the right.

lazarusA:
>  1st to 2nd row should be more to the right.
@storopoli you mean like in the centre? And without the grids I don't like it, it looks too empty. Do a print of both, so maybe you will see what I mean.

lazarusA:
Like this? (grid lines more opaque, I think they should be there... @rikhuijzer ?)
test_front

storopoli:
But in empty images? like in `fig[1, 2]` and `fig[2, 3]`

storopoli:
I still think they should go. It's a cover not a formal "plot"  where you need them

rikhuijzer:
I'm with @lazarusA here. Without, it looks a bit too empty even on thumbnail sizes.

lazarusA:
@rikhuijzer wait... revert the merge please.

PR #144, opened 2021-09-29
Rewrite Julia Basics

rikhuijzer:
I combined some time here and there to read through the section.

### DIFF
--- a/contents/julia_basics.md
+++ b/contents/julia_basics.md
@@ -985,11 +985,28 @@ Speaking in arrays, let's talk about them.
 
 ### Array {#sec:array}
 
-Arrays are a **systematic arrangement of similar objects, usually in _rows_ and _columns_**.
-Most of the time you would want **arrays of a single type for performance issues**, but note that they can also hold objects of different types.
-They are the "bread and butter" of data scientist, because arrays are what constitutes most of **data manipulation** and **data visualization** workflows.
+In its most basic form, arrays hold multiple objects.
+For example, they can hold multiple numbers in one-dimension:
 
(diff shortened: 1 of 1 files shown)


contents/julia_basics.md
```diff
@@ -1037,196 +1051,175 @@ We also have some **syntax aliases** for the most common elements in array const
...
     ```jl
-    sco(
-    """
-    my_vector_zeros = zeros(10)
-    """
-    )
+    s = """
+        my_vector_zeros = zeros(10)
+        """
+    sco(s)
```
rikhuijzer, line 1057:
```suggestion
    sco("my_vector_zeros = zeros(10)")
```
Would have been better, but I forgot to do it here. Maybe next time. Later in the PR, I did go for the shorter form.

storopoli (commented):
Ok, so the idea is to remove the initial bullet points to a more narrative approach?

contents/julia_basics.md
storopoli: There are several like this in the file. Doesn't make sense to change only one...

rikhuijzer:
> Ok, so the idea is to remove the initial bullet points to a more narrative approach?

No. I didn't remove any pullet point. I only did did a rewrite of some small sentences, added some sentences in places where I got confused and I did rewrite
```
sco("""
    foo
    """
)
```
to either
```
s = """
     foo
     """
sco(s)
```
or
```
sco("foo")
```

Let me know if I should revert anything

storopoli (approved):
Ok, thanks! The improvements makes the text better and with the "face" of the three of us. Instead of just one author.

rikhuijzer:
Thank you for your review, Jose!

Issue #17, opened 2021-06-27
Launch the Printed Book at Amazon.com


storopoli:
We need to:

- [x] Normalize functions and syntax to blue codestyle, also accordingly to DataFrames, Plots, Makie, AoG docs
- [x] Create a PDF with syntax highlighting (#15)
- [x] Create an Author Profile at Amazon.com (both @storopoli, @rikhuijzer  and @lazarusA )
- [x] Get a ISBN (Amazon Provides a [free one](https://kdp.amazon.com/en_US/help/topic/G201834170))
- [x] Adapt the book to Amazon.com format
- [x] Create cover (#148) 
- [x] Write a summary
- [x] Tag a "first edition" release on this repository

rikhuijzer:
Just discussed with Lazaro. Size is gonna be 6.14" x 9.21" (15.6 x 23.39 cm). Available sizes: https://kdp.amazon.com/en_US/help/topic/G201834180.

rikhuijzer:
For 230 pages, the cover will look as follows (https://kdp.amazon.com/en_US/cover-templates):

![image](https://user-images.githubusercontent.com/20724914/134786525-d8835fe2-09bf-444b-8897-bf850649cae4.png)

I plan to make an automated build soon for the cover via either Makie or LaTeX.

rikhuijzer:
Which paper?

![image](https://user-images.githubusercontent.com/20724914/135903049-7f94b50c-54ce-45ab-88c3-4718e5f4a686.png)

versus

![image](https://user-images.githubusercontent.com/20724914/135903126-9bd6c148-ddfe-4115-8e47-b3d73642a902.png)

I'd say the cheaper one to make the book more easily available.

rikhuijzer:
Book margins are approved now after https://github.com/rikhuijzer/Books.jl/commit/ce1e8013f7e3928b1ce32ab8dac77f46e84da961:

![image](https://user-images.githubusercontent.com/20724914/135915045-4e64091d-6b44-43a9-b55e-9f60ea6b50d6.png)

storopoli:
What about the price difference from the standard vs premium?

PR #148, opened 2021-10-04
Add full cover

rikhuijzer:
This PR add code to generate the full cover. See https://github.com/JuliaDataScience/JuliaDataScience/issues/17#issuecomment-927185934 for more information about the dimensions.

I don't consider the cover done and perfect now. It is meant as a place where we can start discussing the appearance.

# Preview

image

**EDIT:** This wasn't working in Amazon. It was turned into 

image

So, instead using the Amazon cover editor:

image

which is turned into

image

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -30,7 +30,6 @@ Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StatsPlots = "f3b207a7-027a-5e70-b257-86293d7955fd"
-Tectonic = "9ac5f52a-99c6-489f-af81-462ef484790f"
 TestImages = "5e47fb64-e119-507b-a336-dd2b206d9990"
 XLSX = "fdbf4ff8-1666-58a4-91e7-1b58723a45e0"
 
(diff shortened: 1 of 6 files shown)


storopoli (approved):
Wow!! You've included our favicon in the spine!!!!

lazarusA:
I must say, that now that I see it again.. I do like those scatter points in 3d with colors and everything.... and then probably it will look even better with the volume plot at the top... which is only available via GLMakie. Nice Job @rikhuijzer, it indeed looks great already.

rikhuijzer:
@storopoli can you review this again? I've not included `cover.jl`, but still add the file to have a backup of the code in Git.

cormullion:
_unsolicited suggestion for tweaking cover design_

Screenshot 2021-10-05 at 10 07 20

storopoli:
😂 Amazing we were debating whether to ask for your suggestion. @lazarusA and @rikhuijzer what you think? Also @cormullion please speak freely about the cover. We really value your opinion!

rikhuijzer:
haha yes awesome. Something is still off with the font, I think. But the placing of the text is much better indeed

lazarusA:
@cormullion nice suggestion. if you have a different idea for the cover, we will be happy to hear about it.

rikhuijzer:
Okay, Makie.jl isn't the best tool for the job, but I think it is better now:

image

@lazarusA @storopoli let's merge? We can fine tune it later

lazarusA:
yes.

cormullion:
I'm a fan of Barlow - open source, lots of good heavy weights for titles etc... 

It's useful to check out the idea by viewing the file as a thumbnail. If it works at that size it's probably a good design...

Screenshot 2021-10-05 at 11 15 10

rikhuijzer:
This PR is starting more and more to suffer from scope creep. I think that the current PR is a strict improvement over the previous state, so I'll merge.

The implementation of the nice suggestions by cormulion can be further discussed at https://github.com/JuliaDataScience/JuliaDataScience/issues/151.

PR #152, opened 2021-10-05
Added leftjoin! to performance section

### LINKED ISSUE #149: leftjoin! inplace
We need to include `leftjoin!`.

It will be added in 1.3 release of `DataFrames.jl` and already is in `main` branch.

> EDIT: We only cover non-allocating functions (the bang `!` ones) in performance:

- [x] Performance Section

Source code here: https://github.com/JuliaData/DataFrames.jl/blob/main/src/join/inplace.jl

PS: Thanks for the warning @bkamins!

storopoli:
This PR modifies the performance section of the `DataFrames.jl` chapter.

Since version 1.3 `DataFrames.jl` will support an *in-place* `leftjoin!`.

Thanks for the advice @bkamins.

Closes #149

### DIFF
--- a/contents/dataframes_performance.md
+++ b/contents/dataframes_performance.md
@@ -14,7 +14,7 @@ For example, `filter` has an _in-place_ `filter!`, `select` has `select!`, `subs
 Notice that these functions **do not** return a new `DataFrame`, but instead they **update** the `DataFrame` that they act upon.
 Additionally, `DataFrames.jl` (version 1.3 onwards) supports in-place `leftjoin` with the function `leftjoin!`.
 This function updates the left `DataFrame` with the joined columns from the right `DataFrame`.
-There is a caveat that for each row of left table there must be *at most* one row in right table.
+There is a caveat that for each row of left table there must match *at most* one row in right table.
 
 If you want the highest speed and performance in your code, you should definitely use the `!` functions instead of regular `DataFrames.jl` functions.
 
--- a/contents/dataframes_performance.md
+++ b/contents/dataframes_performance.md
@@ -12,7 +12,9 @@ In the context of high performance Julia code, this *means* that **functions wit
 Almost all the `DataFrames.jl` functions that we've seen have a \"`!` twin\".
 For example, `filter` has an _in-place_ `filter!`, `select` has `select!`, `subset` has `subset!`, and so on.
 Notice that these functions **do not** return a new `DataFrame`, but instead they **update** the `DataFrame` that they act upon.
-Additionally, `DataFrames.jl` (version 1.3 onwards) supports in-place `leftjoin` with the function `leftjoin!` by updating the left `DataFrame` with the joined columns from right `DataFrame`.
+Additionally, `DataFrames.jl` (version 1.3 onwards) supports in-place `leftjoin` with the function `leftjoin!`.
+This function updates the left `DataFrame` with the joined columns from the right `DataFrame`.
+There is a caveat that for each row of left table there must be *at most* one row in right table.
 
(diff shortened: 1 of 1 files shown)


contents/dataframes_performance.md
```diff
@@ -12,9 +12,7 @@ In the context of high performance Julia code, this *means* that **functions wit
 Almost all the `DataFrames.jl` functions that we've seen have a \"`!` twin\".
 For example, `filter` has an _in-place_ `filter!`, `select` has `select!`, `subset` has `subset!`, and so on.
 Notice that these functions **do not** return a new `DataFrame`, but instead they **update** the `DataFrame` that they act upon.
-Additionally, there are some functions that do not have a `!` counterpart.
-For example, all the `join`s, for technical reasons, *cannot* be done in-place.
-Thus, we don't have any `join!` functions in `DataFrames.jl`.
+Additionally, `DataFrames.jl` (version 1.3 onwards) supports in-place `leftjoin` with the function `leftjoin!` by updating the left `DataFrame` with the joined columns from right `DataFrame`.
```
rikhuijzer, line 15:
```suggestion
Additionally, `DataFrames.jl` (version 1.3 onwards) supports in-place `leftjoin` with the function `leftjoin!`.
This function updates the left `DataFrame` with the joined columns from the right `DataFrame`.
```

rikhuijzer (approved):

contents/dataframes_performance.md
rikhuijzer: Upon looking at this PR, I was like thinking whether we can say a bit more about it. However, the API is completely in line with all the other functions from DataFrames.jl and I don't see any reason for surprises in the implementation. So, yes this should be enough.

bkamins: Two comments:
* in general right table can be just data frame (not only `DataFrame`, but also `SubDataFrame`), but it is a detail - so you might want to keep things as is as this is simpler
* the additional requirement for `leftjoin!` is that for each row of left table there must be **at most** one row in right table; the issue is that if we do in-place update we do not want to produce duplicate matches as then they would mean we would have to increase the number of rows in left data frame; in normal `leftjoin` having duplicates is allowed.

storopoli: We are trying not to go into `SubDataFrame` in the first edition of the book. But we will cover in a future edition/version.

storopoli:
So if you guys are OK, we can merge it.

contents/dataframes_performance.md
```diff
@@ -12,9 +12,9 @@ In the context of high performance Julia code, this *means* that **functions wit
 Almost all the `DataFrames.jl` functions that we've seen have a \"`!` twin\".
 For example, `filter` has an _in-place_ `filter!`, `select` has `select!`, `subset` has `subset!`, and so on.
 Notice that these functions **do not** return a new `DataFrame`, but instead they **update** the `DataFrame` that they act upon.
-Additionally, there are some functions that do not have a `!` counterpart.
-For example, all the `join`s, for technical reasons, *cannot* be done in-place.
-Thus, we don't have any `join!` functions in `DataFrames.jl`.
+Additionally, `DataFrames.jl` (version 1.3 onwards) supports in-place `leftjoin` with the function `leftjoin!`.
+This function updates the left `DataFrame` with the joined columns from the right `DataFrame`.
+There is a caveat that for each row of left table there must be *at most* one row in right table.
```
bkamins, line 17:
```suggestion
There is a caveat that for each row of left table there must match *at most* one row in right table.
```

storopoli:
Thanks! I am merging it!

bkamins (approved):

PR #153, opened 2021-10-05
Better fig sizes makie

lazarusA:
I added some new fig sizes since in the previous version the pdf output is not that good. This one is a small improvement :D


rikhuijzer:
Could you give a high level description of all the changes in this PR? I'm a bit confused :)

PR #156, opened 2021-10-06
[Stats] Fix Stats Definition

storopoli:
This PR fixes the definition to probability and removes the old definition that seeded discord.
I also added something to the "classical definition of probability" to not get into discord with measure theory also.

### DIFF
--- a/contents/stats_distributions.md
+++ b/contents/stats_distributions.md
@@ -3,7 +3,7 @@
 Before defining distributions, we must talk about **probability**.
 
 **The probability of an event is a real number $p$ where $0 \leq p \leq 1$.
-Here, 0 indicates the that an event is impossible and 1 indicates that an event will certainly happen.**
+Here, 0 indicates the that an event will [almost never](https://en.wikipedia.org/wiki/Almost_surely) occur and 1 indicates that an event will [almost surely] (https://en.wikipedia.org/wiki/Almost_surely).**
 The greater the likelihood of an event, the more likely it is that the event will occur.
 A simple example is the tossing of a fair (impartial) coin.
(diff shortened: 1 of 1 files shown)


contents/stats_distributions.md
```diff
@@ -2,16 +2,12 @@
 
 Before defining distributions, we must talk about **probability**.
 
-**The probability of an event is a real number between 0 and 1, where 0 indicates the impossibility of the event and 1 indicates the certainty of the event**.
+In the classical definition of probability, **the probability of an event is a real number between 0 and 1, where 0 indicates the impossibility of the event and 1 indicates the certainty of the event**.
```
rikhuijzer, line 5:
```suggestion
**The probability of an event is a real number $p$ where $0 \leq p \leq 1$.
Here, 0 indicates the that an event is impossible and 1 indicates that an event will certainly happen.**
```
One of the critiques was that "between 0 and 1" is not the same as "the closed interval between 0 and 1", that is, $[a, b] = \{ x \in \mathbb{R} | a \leq x \leq b \}$.

Open and closed intervals cannot be expressed clearly in the English language, so that's why I suggest using $x$ where $a \leq x \leq b$ because that notation requires less knowledge from the reader than $[a, b]$.

rikhuijzer (changes requested):

rikhuijzer (approved):

rikhuijzer:
I'm also okay with "classical definition of probability", but didn't read your justification. Only just now I read it.

spinkney (commented):
0 is not impossible and 1 is not a certainty see definitions of "almost never" and "almost always" https://en.wikipedia.org/wiki/Almost_surely. It's why on continuous spaces that P(A <= a) = P(A < a) since P(A == a) = 0 but "a" can still be in the set of possibilities, thus it's possible, we just don't expect it to happen.

storopoli:
Thanks @spinkney! I've updated the definition.

rikhuijzer:
I think I have a solution for these discussions in #158.

PR #161, opened 2021-10-07
Corrected discrete CDF plot

storopoli:
This is a fix for the CDF plot of a Discrete Uniform random variable (the dice example) that was as a line and now is a step-wise function.
image

This was spotted by @mschauer, thanks!

@lazarusA will probably want to tweak something in Makie (I don't think I have the most elegant solution).


lazarusA:
I don't see why your ticks start on the left.... they should be in the centre at least isn't? If, I think for instance on the last case... does that mean that one goes from 6 to 7 ? Also, all plots should have labels. (if not possible then the corresponding legend).

PR #165, opened 2021-10-09
Show where trailing whitespace occurs if found

rikhuijzer:
Very minor change. When trailing whitespace is found, the GitHub Actions log will now show which lines cause the format check to fail.


storopoli:
I've updated the JuliaFormatter compat to 0.17. Is that ok?

PR #171, opened 2021-10-13
Add `using LaTeXStrings`

rikhuijzer:
I was trying to use LaTeX equations in Makie with our book, but it wasn't completely clear that it is required to first load `LaTeXStrings.jl`.


lazarusA:
Makie already loads the basic `L` but not `latexstring`, so `using LaTeXString` is only necessary for more complicated examples. And, yes I agree with your suggestion, but then we need to remove the line after the paragraph shown above.

PR #185, opened 2021-10-21
Fix Shell Files in .gitattributes

storopoli:
This is annoying:
image


rikhuijzer:
> This is annoying

:rofl: 

:100:

PR #197, opened 2021-10-31
Typo in `naive_trouble`: chicken & fox

vmikk:
Hello!
Probably there is a typo in the `naive_trouble` function.
The second and third conditions are identical (both, A and B are chickens):
```
  function naive_trouble(A::Animal, B::Animal)
        if A isa Fox && B isa Chicken
            return true
        elseif A isa Chicken && B isa Chicken
            return true
        elseif A isa Chicken && B isa Chicken
            return false
        end
    end
```

I assume that in the second conditions B should be a Fox, isn't it?
```
  function naive_trouble(A::Animal, B::Animal)
        if A isa Fox && B isa Chicken
            return true
        elseif A isa Chicken && B isa Fox
            return true
        elseif A isa Chicken && B isa Chicken
            return false
        end
    en
```
With kind regards,
Vladimir

### DIFF
--- a/contents/why_julia.md
+++ b/contents/why_julia.md
@@ -228,7 +228,7 @@ s = """
     function naive_trouble(A::Animal, B::Animal)
         if A isa Fox && B isa Chicken
             return true
-        elseif A isa Chicken && B isa Chicken
+        elseif A isa Chicken && B isa Fox
             return true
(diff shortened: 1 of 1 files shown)


rikhuijzer:
Awesome. Well spotted. Thanks!

PR #199, opened 2021-11-03
Add Pirsch analytics

rikhuijzer:
@lazarusA is this okay? Pirsch offers the same functionality as Fathom and is also very private by design, but they do it for a much lower price. If you agree, I plan to remove Fathom in a month. For now, they can run side by side.


lazarusA:
yes is ok.

lazarusA (approved):

PR #208, opened 2021-11-24
Remove Fathom analytics

rikhuijzer:
This will also slightly speed up page loading speed. Unless any of you disagrees, I'll remove Fathom and its data in a few days.


storopoli (approved):
It already has passed +/- 1 month to make the change from Fathom to Pirsch

Accoring to #199

PR #212, opened 2021-11-29
typo fix dataframes

NMLebedev:
Hello, I'm Nadia. I'm translating your book to portuguese, with @storopoli help.

I think this should be were, not where.


rikhuijzer:
Well spotted. Thank you, Nadia!

PR #214, opened 2021-12-04
Update julia_basics.md

rcqls:
This tiny change seems for me more natural. Thanks for this very nice book!


storopoli (approved):
Thank you!

The help prompt for the function are:

```julia
startswith(s::AbstractString, prefix::AbstractString)
endswith(s::AbstractString, suffix::AbstractString)
```

So you are absolutely right! Thanks!

PR #217, opened 2021-12-05
Julia Basics: improvements and suggestions

Mo-Gul:
feel free to cherry-pick

### DIFF
--- a/contents/julia_basics.md
+++ b/contents/julia_basics.md
@@ -1849,13 +1849,13 @@ Whenever you are dealing with files such as CSV, Excel files or other Julia scri
 This is easily accomplished with the `joinpath`, `@__FILE__` and `pkgdir` functions.
 
 If you write your code in a package, you can use `pkgdir` to get the root directory of the package.
-For example, for the Julia Data Science (JDS) package that we use to produce this book:
+For example, for the Julia Data Science (JDS) package that we use to produce this book the root directory is
 
 ```jl
 root = pkgdir(JDS)
 ```
 
-as you can see, the code to produce this book was running on a Linux computer.
+As you can see, the code to produce this book was running on a Linux computer.
 If you're using a script, you can get the location of the script file via
 
 ```julia
--- a/contents/julia_basics.md
+++ b/contents/julia_basics.md
@@ -669,7 +669,6 @@ sco("logarithm.([1, 2, 3])")
 ```
 
 #### Functions with a bang `!` {#sec:function_bang}
-<!-- NOTE: make it an unnumbered section (there is no section 3.3.1.2) -->
 
 It is a Julia convention to append a bang `!` to names of functions that modify one or more of their arguments.
 This convention warns the user that the function is **not pure**, i.e., that it has *side effects*.
@@ -1789,7 +1788,7 @@ We use `Symbol`s a lot in data manipulations with the `DataFrames.jl` package (@
 ### Splat Operator {#sec:splat}
 
 In Julia we have the "splat" operator `...` which is used in function calls as a **sequence of arguments**.
-We will occasionally use splatting in some function calls in the **data manipulation** and **data visualization** sections.
+We will occasionally use splatting in some function calls in the **data manipulation** and **data visualization** chapters.
 
 The most intuitive way to learn about splatting is with an example.
 The `add_elements` function below takes three arguments to be added together:
@@ -1839,7 +1838,7 @@ We share code, data, tables, figures and so on.
 Behind everything, there is the **operating system (OS) filesystem**.
 In a perfect world, the same program would give the **same** output when running on **different** operating systems.
 Unfortunately, that is not always the case.
-One instance of this is the difference between Windows paths, such as `C:\user\john`, and Linux paths, such as `/home/john`.
+One instance of this is the difference between Windows paths, such as `C:\\user\john`, and Linux paths, such as `/home/john`.
 This is why it is important to discuss **filesystem best practices**.
 
 Julia has native filesystem capabilities that **handle the differences between operating systems**.
@@ -1849,7 +1848,7 @@ Whenever you are dealing with files such as CSV, Excel files or other Julia scri
 This is easily accomplished with the `joinpath`, `@__FILE__` and `pkgdir` functions.
 
 If you write your code in a package, you can use `pkgdir` to get the root directory of the package.
-For example, for the Julia Data Science (JDS) package that we use to produce this book the root directory is
+For example, for the Julia Data Science (JDS) package that we use to produce this book, the root directory is:
 
 ```jl
 root = pkgdir(JDS)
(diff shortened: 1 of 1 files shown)


storopoli:
@Mo-Gul thanks for the contribution. Regarding the section things I will ignore them because they are much more related to [`Books.jl`](https://github.com/JuliaBooks/Books.jl).

contents/julia_basics.md
```diff
@@ -669,6 +669,7 @@ sco("logarithm.([1, 2, 3])")
 ```
 
 #### Functions with a bang `!` {#sec:function_bang}
+<!-- NOTE: make it an unnumbered section (there is no section 3.3.1.2) -->
```
storopoli, line 672:
```suggestion
```

contents/julia_basics.md
```diff
@@ -1788,7 +1789,7 @@ We use `Symbol`s a lot in data manipulations with the `DataFrames.jl` package (@
 ### Splat Operator {#sec:splat}
 
 In Julia we have the "splat" operator `...` which is used in function calls as a **sequence of arguments**.
-We will occasionally use splatting in some function calls in the **data manipulation** and **data visualization** chapters.
+We will occasionally use splatting in some function calls in the **data manipulation** and **data visualization** sections.
```
storopoli, line 1792:
```suggestion
We will occasionally use splatting in some function calls in the **data manipulation** and **data visualization** chapters.
```

contents/julia_basics.md
```diff
@@ -1838,7 +1839,7 @@ We share code, data, tables, figures and so on.
 Behind everything, there is the **operating system (OS) filesystem**.
 In a perfect world, the same program would give the **same** output when running on **different** operating systems.
 Unfortunately, that is not always the case.
-One instance of this is the difference between Windows paths, such as `C:\\user\john\`, and Linux paths, such as `/home/john`.
+One instance of this is the difference between Windows paths, such as `C:\user\john`, and Linux paths, such as `/home/john`.
```
storopoli, line 1842:
```suggestion
One instance of this is the difference between Windows paths, such as `C:\\user\john`, and Linux paths, such as `/home/john`.
```

Windows uses the double inverted slash bar thing before a drive.

contents/julia_basics.md
```diff
@@ -1848,13 +1849,13 @@ Whenever you are dealing with files such as CSV, Excel files or other Julia scri
 This is easily accomplished with the `joinpath`, `@__FILE__` and `pkgdir` functions.
 
 If you write your code in a package, you can use `pkgdir` to get the root directory of the package.
-For example, for the Julia Data Science (JDS) package that we use to produce this book:
+For example, for the Julia Data Science (JDS) package that we use to produce this book the root directory is
```
storopoli, line 1852:
```suggestion
For example, for the Julia Data Science (JDS) package that we use to produce this book, the root directory is:
```

Missing a coma and colon

contents/julia_basics.md
```diff
@@ -2350,7 +2351,7 @@ By default, it returns a string that holds the file path for the downloaded file
 s = """
     url = "https://raw.githubusercontent.com/JuliaDataScience/JuliaDataScience/main/Project.toml"
 
-    my_file = Downloads.download(url) # tempfile() being created
+    my_file = download(url) # tempfile() being created
```
storopoli, line 2354:
```suggestion
    my_file = Downloads.download(url) # tempfile() being created
```

We need this because of conflicts with `Base.download`. See the failing GH actions...

storopoli (changes requested):

contents/julia_basics.md
Mo-Gul: Sorry, but I was not sure if you just wanted to give a path as would normally be given/shown in the Windows explorer or `cmd`/`powershell`. There would be only *one* backslash each. When I do e.g.

```julia
root = dirname(@__FILE__)
```

in my Julia test file I get double backlashes *everywhere*, i.e. not only after the drive letter. So I e.g. get `d:\\my\\path\\to\\Julia\\test.jl`.

So when the intent was to show a valid Windows path in Julia the above instance should be changed to

```md
... such as `C:\\user\\john`, ...
```

contents/julia_basics.md
Mo-Gul: But when I do (in Windows using Julia v1.7)

```julia
using Downloads: download
url = "https://raw.githubusercontent.com/JuliaDataScience/JuliaDataScience/main/Project.toml"
my_file = Downloads.download(url)
```

I get `ERROR: LoadError: UndefVarError: Downloads not defined`. 

It seems to work if I just use

```julia
using Downloads
```

contents/julia_basics.md
rikhuijzer: Those double backslashes aren't always shown :)

```julia
julia> raw"foo\bar"
"foo\\bar"

julia> raw"foo\bar" |> print
foo\bar
```

contents/julia_basics.md
rikhuijzer: This should work:
```
using Downloads: download
url = "https://raw.githubusercontent.com/JuliaDataScience/JuliaDataScience/main/Project.toml"
my_file = download(url)
```
Further discussion about `using Downloads` versus `using Downloads: download` is in the notation section (https://juliadatascience.io/notation). Specifically, see `6.2.3.1 Loading of symbols`.

Mo-Gul: @rikhuijzer, I agree. That was my first suggestion (as you can see above) which then was overruled by @storopoli. So it seems both of you have to finally discuss what is "right"/better ...

contents/julia_basics.md
Mo-Gul: @rikhuijzer, (as a reminder, I have just started with Julia) and I can see your argument. But can you image an output (of `dirname(@__FILE__) on Windows) where you will get a **mix** of single and double backslashes? Because that is what currently is written in the book ...

My suggestion was and is to show either **all** instances of backslashes as "singles" or as "doubles" (but not as a mixture of both).

contents/julia_basics.md
storopoli: @Mo-Gul sorry, you might be right here. But could you please make a PR with a not-failing GH Actions CI.

I think you'll need to change the https://github.com/JuliaDataScience/JuliaDataScience/blob/main/src/JDS.jl#L52 to:

```julia
using Downloads: download
```

Mo-Gul: @storopoli, sorry, but I leave this for you. I don't want to mess around with that stuff.

I would (in the meantime) continue *reading* the book after some days of filing issues and preparing PRs ;)

contents/julia_basics.md
rikhuijzer: Fixed in https://github.com/JuliaDataScience/JuliaDataScience/pull/224.

contents/julia_basics.md
rikhuijzer: See https://github.com/JuliaDataScience/JuliaDataScience/pull/223.

PR #218, opened 2021-12-06
improve PR #214

Mo-Gul:
while the syntax of the commands `startswith` and `endswith` was corrected in PR #214, the surrounding text wasn't adapted to that change as well. This is done in this PR. To simplify that, also the command `occursin` was changed to `contains` which has the same syntax as `startswith` and `endwith` (and internally calls `occursin`).

### DIFF
--- a/contents/julia_basics.md
+++ b/contents/julia_basics.md
@@ -818,20 +818,20 @@ scob(s)
 ```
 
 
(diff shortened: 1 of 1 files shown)


storopoli:
Thanks, are you sure that `occursin` do not appear anywhere else in the book?

storopoli (approved):

Mo-Gul:
Yes, I am sure.

PR #223, opened 2021-12-08
Fix downloads

storopoli:
This is an issue that @Mo-Gul has brought up and we are now creating a new PR to address it.

### DIFF
--- a/contents/julia_basics.md
+++ b/contents/julia_basics.md
@@ -2327,7 +2327,7 @@ sco("rand(my_seed, 3)")
 
 ### Downloads {#sec:downloads}
 
-One last thing from Julia's standard library for us to cover is the **`Download` module**.
+One last thing from Julia's standard library for us to cover is the **`Downloads` module**.
 It will be really brief because we will only be covering a single function named `download`.
 
 Suppose you want to **download a file from the internet to your local storage**.
@@ -2336,21 +2336,20 @@ The first and only required argument is the file's url.
 You can also specify as a second argument the desired output path for the downloaded file (don't forget the filesystem best practices!).
 If you don't specify a second argument, Julia will, by default, create a temporary file with the `tempfile` function.
 
-Let's load the `download` method:
+Let's load the `download` method from the `Downloads` module:
 
 ```julia
 using Downloads: download
 ```
 
 For example, let's download our [`JuliaDataScience` GitHub repository](https://github.com/JuliaDataScience/JuliaDataScience) `Project.toml` file.
-Note that `download` function is not exported by `Downloads` module, so we have to use the `Module.function` syntax.
 By default, it returns a string that holds the file path for the downloaded file:
 
 ```jl
 s = """
     url = "https://raw.githubusercontent.com/JuliaDataScience/JuliaDataScience/main/Project.toml"
 
-    my_file = Downloads.download(url) # tempfile() being created
+    my_file = download(url) # tempfile() being created
     """
 scob(s)
 ```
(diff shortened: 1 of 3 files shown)


storopoli:
@rikhuijzer, I cannot get it to pass GH Actions. Do you have any insights?

lazarusA:
@storopoli  we have a name conflict here: I think that's why I was used it like that: 

```
WARNING: both JDS and Base export "download"; uses of it in module Main must be qualified
255
ERROR: UndefVarError: download not defined
```

src/JDS.jl
```diff
@@ -69,6 +69,7 @@ using Reexport: @reexport
 end # @reexport
 
 const SMALL_IM_ATTR = "width=70%"
+const download = Downloads.download
```
rikhuijzer, line 72: This is hidden to the reader. Maybe we should just keep it as it used to be? The suggestion to use `using Foo: bar` is a suggestion and not meant to be a strict rule.

storopoli: This is just the intent, this is hidden. I am trying to make the:

```julia
using Downloads: download

url = "https://raw.githubusercontent.com/JuliaDataScience/JuliaDataScience/main/Project.toml"
my_file = download(url) # tempfile() being created
```

in `julia_basics.md` to work.

rikhuijzer: That will not work for the reader either. In Julia 1.6, there are `Base.download` and `Download.download`:
```
help?> Base.download
  download(url::AbstractString, [path::AbstractString = tempname()]) -> path

  Download a file from the given url, saving it to the location path, or if not specified, a temporary path. Returns the path of the downloaded file.

  │ Note
  │
  │  Since Julia 1.6, this function is deprecated and is just a thin wrapper around Downloads.download. In new code, you should use that function directly instead
  │  of calling this.
```

src/makie.jl
```diff
@@ -592,7 +592,7 @@ function mixing_surface_contour3d_contour_and_contourf()
 end
 
 url = "https://raw.githubusercontent.com/JuliaImages/TestImages.jl/images/images/monarch_color_256.png"
-img = load(Downloads.download(url))
+img = load(download(url))
```
storopoli, line 595:
```suggestion
img = load(Downloads.download(url))
```

src/JDS.jl
storopoli:
```suggestion
```

contents/julia_basics.md
```diff
@@ -2336,21 +2336,20 @@ The first and only required argument is the file's url.
 You can also specify as a second argument the desired output path for the downloaded file (don't forget the filesystem best practices!).
 If you don't specify a second argument, Julia will, by default, create a temporary file with the `tempfile` function.
 
-Let's load the `download` method:
+Let's load the `download` method from the `Downloads` module:
```
storopoli, line 2339:
```suggestion
Let's load the `Downloads` module:
```

contents/julia_basics.md
```diff
@@ -2336,21 +2336,20 @@ The first and only required argument is the file's url.
...
 
 ```julia
 using Downloads: download
```
storopoli, line 2342:
```suggestion
using Downloads
```

contents/julia_basics.md
```diff
@@ -2336,21 +2336,20 @@ The first and only required argument is the file's url.
...
 ```
 
 For example, let's download our [`JuliaDataScience` GitHub repository](https://github.com/JuliaDataScience/JuliaDataScience) `Project.toml` file.
```
storopoli, line 2345:
```suggestion
For example, let's download our [`JuliaDataScience` GitHub repository](https://github.com/JuliaDataScience/JuliaDataScience) `Project.toml` file.
Note that `download` function is not exported by `Downloads` module, so we have to use the `Module.function` syntax.
```

storopoli:
Ok, I think now it will work.

rikhuijzer (approved):
Thank you. Looks good to me

PR #225, opened 2021-12-09
Track links on homepage

rikhuijzer:
Someone in our university said that she would immediately click on the PDF link if she would look at our site. With Pirsch, we can luckily also add a onclick event to those links to get a better idea of how many people prefer to look at the PDF instead of HTML.


storopoli (approved):
Perfect!

lazarusA (approved):
Nice!

PR #229, opened 2021-12-11
Remove anonymous

### LINKED ISSUE #227: [dataframes_indexing] Not an anonymous function
From what I have learned so far there is no "anonymous function" in

https://github.com/JuliaDataScience/JuliaDataScience/blob/b5582d29a9afa300e0a124ad2820389c386c04cc/contents/dataframes_indexing.md#L160-L177

When I am correct you declare a function that is then used in `filter`. The `=>` should be the `Pair` operator while `->` is the operator to define an anonymous function, right?

If I am correct, please rephrase the text or adapt the example code.  
If I am wrong, shame on me and I am not a good "student" and sorry for the noise.

rikhuijzer:
@Mo-Gul, well spotted.

Can you double check whether this fix is correct?

- Closes #227

### DIFF
--- a/contents/dataframes_indexing.md
+++ b/contents/dataframes_indexing.md
@@ -157,7 +157,7 @@ s = """filter(:name => !=("Alice"), grades_2020())"""
 sco(s; process=without_caption_label)
 ```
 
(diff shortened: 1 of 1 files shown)


Mo-Gul (approved):
That's fine for me.

PR #230, opened 2021-12-11
Improve data frames chapter

Mo-Gul:
For non-typo changes I added some comments so that you better understand the reason for the (proposed) change.

As always, feel free to cherry-pick.

### DIFF
--- a/contents/dataframes_groupby_combine.md
+++ b/contents/dataframes_groupby_combine.md
@@ -1,5 +1,10 @@
 ## Groupby and Combine {#sec:groupby_combine}
 
+### Single Source Columns
+<!-- That fits to the later "Multiple Source Columns" and also solves the
+     "problem" stated in issue #221.
+     (Seems like I have missed this one too ...) -->
+
 In the R programming language, @wickham2011split has popularized the so-called split-apply-combine strategy for data transformations.
 In essence, this strategy **splits** a dataset into distinct groups, **applies** one or more functions to each group, and then **combines** the result.
 `DataFrames.jl` fully supports split-apply-combine.
--- a/contents/dataframes_groupby_combine.md
+++ b/contents/dataframes_groupby_combine.md
@@ -1,10 +1,5 @@
 ## Groupby and Combine {#sec:groupby_combine}
 
-### Single Source Columns
-<!-- That fits to the later "Multiple Source Columns" and also solves the
-     "problem" stated in issue #221.
-     (Seems like I have missed this one too ...) -->
-
 In the R programming language, @wickham2011split has popularized the so-called split-apply-combine strategy for data transformations.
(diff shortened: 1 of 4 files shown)


contents/dataframes_groupby_combine.md
```diff
@@ -1,5 +1,10 @@
 ## Groupby and Combine {#sec:groupby_combine}
 
+### Single Source Columns
+<!-- That fits to the later "Multiple Source Columns" and also solves the
+     "problem" stated in issue #221.
+     (Seems like I have missed this one too ...) -->
+
```
rikhuijzer, line 7:
```suggestion
```
It's unclear what "problem" you are referring too. I prefer to have text at the start of each section and not start a section immediately with a subsection.

contents/dataframes_transform.md
```diff
@@ -4,6 +4,11 @@
 We need to cover `ifelse` and `case_when`
 ```
 
+### Single Transformations
+<!-- That fits to the later "Multiple Transformation" and also solves the
+     "problem" stated in issue #221.
+     (Seems like I have missed this one ...) -->
+
```
rikhuijzer, line 11:
```suggestion
```

Same comment as above

contents/dataframes_transform.md
```diff
@@ -16,8 +21,15 @@ In `DataFrames.jl`, the syntax is `source => transformation => target`.
 Like before, we use the `grades_2020` dataset:
 
 ```jl
-@sco process=without_caption_label grades_2020()
+s = "grades_2020()"
```
rikhuijzer, line 25:
```suggestion
@sco process=without_caption_label grades_2020()
```

contents/dataframes_transform.md
```diff
@@ -16,8 +21,15 @@ In `DataFrames.jl`, the syntax is `source => transformation => target`.
...
 ```
+<!-- It is not necessary to show the definition of the function here again.
+     This is especially true because the definition of `grades_2021` was
+     not/never shown in the previous section.
+     I used that as an exercise to create it on my own.
+     In case I wouldn't have been able to do that, I would have peeked at
+     the definition of `grades_2020` :-) -->
```
rikhuijzer, line 32:
```suggestion
```
Lengthy comments like this can be put in GitHub issue/PR so that they don't end up in the file.

Mo-Gul: I know that this *would* be possible. Unfortunately than my work flow would be much more complicated, because I make these comments directly when reading. The comments I would have to add *after*
- I have read the chapter/section,
- made the commits, and
- made the PR

So please forgive me when I'll continue adding comments to the code.  
(The other option would be to *not* add these comments at all, but then I guess some of the stuff that I suggest would not be self-explanatory ...)

PR #235, opened 2021-12-15
Update rand text

### LINKED ISSUE #234: Julia cannot reproduce the rand
I have put an issue on Stack Overflow about an example of this book. Could you please explain? https://stackoverflow.com/questions/70321085/julia-cannot-reproduce-the-rand

Many thanks.

Shixiang

rikhuijzer:
Fixes #234.

This was caused by a collision of the output filenames of the blocks again. I've added some comments to fix that problem.

EDIT: Also, I've simplified the text a bit.

### DIFF
--- a/contents/julia_basics.md
+++ b/contents/julia_basics.md
@@ -2224,7 +2224,7 @@ scob("rand(1.0:10.0)")
 ```
 
 You can also specify a different step size inside the interval and a different type.
-Here we are using numbers without the dot `.` so Julia will interpret them as `Int64` and not `Float64`:
+Here we are using numbers without the dot `.` so Julia will interpret them as `Int64` and not as `Float64`:
 
 ```jl
 scob("rand(2:2:20)")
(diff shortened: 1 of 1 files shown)


contents/julia_basics.md
```diff
@@ -2228,7 +2224,7 @@ scob("rand(1.0:10.0)")
 ```
 
 You can also specify a different step size inside the interval and a different type.
-Here we are using numbers without the dot `.` so Julia will interpret them as `Int64`:
+Here we are using numbers without the dot `.` so Julia will interpret them as `Int64` and not `Float64`:
```
storopoli, line 2227:
```suggestion
Here we are using numbers without the dot `.` so Julia will interpret them as `Int64` and not as `Float64`:
```

storopoli (approved):
Thank you I saw the stack overflow question and was planning to make the PR sometime next week! ❤️

PR #236, opened 2021-12-20
Improve Makie chapter

Mo-Gul:
As always, feel free to cherry-pick stuff and/or move parts to separate issues in case you don't want to solve this now.

### DIFF
--- a/contents/data_vis_makie.md
+++ b/contents/data_vis_makie.md
@@ -15,9 +15,13 @@ To convert these plot objects to an image, you need one of the Makie backends.
 By default, `Makie.jl` is reexported by every backend, so you only need to install and load the backend that you want to use.
 
 There are three main backends which concretely implement all abstract rendering capabilities defined in Makie.
-One for non-interactive 2D publication-quality vector graphics: `CairoMakie.jl`.
-Another for interactive 2D and 3D plotting in standalone `GLFW.jl` windows (also GPU-powered), `GLMakie.jl`.
-And the third one, a WebGL-based interactive 2D and 3D plotting that runs within browsers, `WGLMakie.jl`. [See Makie's documentation for more](http://makie.juliaplots.org/stable/documentation/backends_and_output/).
+These are
+
+- `CairoMakie.jl` for non-interactive 2D publication-quality vector graphics,
+- `GLMakie.jl` for interactive 2D and 3D plotting in standalone `GLFW.jl` windows (also GPU-powered), and
+- `WGLMakie.jl`, a WebGL-based interactive 2D and 3D plotting that runs within browsers.
+
+[See Makie's documentation for more](http://makie.juliaplots.org/stable/documentation/backends_and_output/).
 
 In this book we will only show examples for `CairoMakie.jl` and `GLMakie.jl`.
 
--- a/contents/data_vis_makie.md
+++ b/contents/data_vis_makie.md
@@ -11,10 +11,10 @@ In our opinion, it is the prettiest and most versatile plotting package.
 Like many plotting packages, the code is split into multiple packages.
 `Makie.jl` is the front end package that defines all plotting functions required to create plot objects.
 These objects store all information about the plots, but still need to be converted to an image.
-To convert these plot objects to an image, you need one of the Makie back ends.
-By default, `Makie.jl` is reexported by every backend, so you only need to install and load the back end that you want to use.
+To convert these plot objects to an image, you need one of the Makie backends.
+By default, `Makie.jl` is reexported by every backend, so you only need to install and load the backend that you want to use.
 
-There are three main back ends which concretely implement all abstract rendering capabilities defined in Makie.
+There are three main backends which concretely implement all abstract rendering capabilities defined in Makie.
 One for non-interactive 2D publication-quality vector graphics: `CairoMakie.jl`.
 Another for interactive 2D and 3D plotting in standalone `GLFW.jl` windows (also GPU-powered), `GLMakie.jl`.
 And the third one, a WebGL-based interactive 2D and 3D plotting that runs within browsers, `WGLMakie.jl`. [See Makie's documentation for more](http://makie.juliaplots.org/stable/documentation/backends_and_output/).
--- a/contents/data_vis_makie_attributes.md
+++ b/contents/data_vis_makie_attributes.md
@@ -118,12 +118,12 @@ sco(s)
 Note that the last line is `set_theme!()`, which will reset to the default settings of Makie.
 For more on `themes` please go to @sec:themes.
 
-Before moving on into the next section, it's worthwhile to see an example where an `array` of attributes are passed at once to a plotting function.
+Before moving on into the next section, it's worthwhile to see an example where an `array` of attributes is passed at once to a plotting function.
 For this example, we will use the `scatter` plotting function to do a bubble plot.
 
-The data for this could be an `array` with 100 rows and 3 columns, here we generated these at random from a normal distribution.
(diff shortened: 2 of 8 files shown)


Mo-Gul:
And here two more remaining questions that I have:

1. In this chapter you used `LinRange` all over the place, but wouldn't it be better (for beginners) to use `range` instead (to avoid stuff that is shown in the example of the [`LinRange` docs](https://docs.julialang.org/en/v1/base/collections/#Base.LinRange)?
2. What is `RGBf0`? (Couldn't find that "colorspace" (or whatever it may be) anywhere so far ...)

And my final thanks (again), because it was the last chapter. Great introduction, although there is left a lot to study in the Makie chapter, because after a smooth start it quickly turns into a showcase demo/gallery. Of course this is just a diagnosis. There *have* to be these showcases where one can find inspirations. And especially they are great when I need/want to do "x" and I remember that there was that example in the gallery where I can look that up.

contents/data_vis_makie.md
```diff
@@ -11,13 +11,17 @@ In our opinion, it is the prettiest and most versatile plotting package.
...
-Another for interactive 2D and 3D plotting in standalone `GLFW.jl` windows (also GPU-powered), `GLMakie.jl`.
-And the third one, a WebGL-based interactive 2D and 3D plotting that runs within browsers, `WGLMakie.jl`. [See Makie's documentation for more](http://makie.juliaplots.org/stable/documentation/backends_and_output/).
+There are three main backends which concretely implement all abstract rendering capabilities defined in Makie.
+These are
+
+- `CairoMakie.jl` for non-interactive 2D publication-quality vector graphics,
+- `GLMakie.jl` for interactive 2D and 3D plotting in standalone `GLFW.jl` windows (also GPU-powered), and
+- `WGLMakie.jl`, a WebGL-based interactive 2D and 3D plotting that runs within browsers.
+
+[See Makie's documentation for more](http://makie.juliaplots.org/stable/documentation/backends_and_output/).
```
lazarusA, line 25: thanks, actually originally the format was like this.

contents/data_vis_makie_attributes.md
```diff
@@ -13,7 +13,7 @@ s = """
 sco(s)
 ```
 
-Or as a `Dict` calling `pltobject.attributes.attributes`.
+Or as a `Dict` calling `pltobj.attributes.attributes`.
```
lazarusA, line 16: nice!

contents/data_vis_makie_attributes.md
```diff
@@ -80,6 +80,14 @@ sco(s)
 Other positions are also available by combining `left(l), center(c), right(r)` and `bottom(b), center(c), top(t)`.
 For instance, for left top, use `:lt`.
 
+<!--
+When I remember correct (I didn't check it now), up to now you really just have
+put stuff in backticks that was code. Now it seem that it is also used to
+emphasize stuff like `legend` or `left(l), center(c), right(r)` and `bottom(b),
+center(c), top(t)`. If it is really just emphasizing and not *also* code, then
```
lazarusA, line 87: yes, only these are like this, but because later they used as keyword arguments. But maybe is better to use a bold face font instead.

contents/data_vis_makie_colors.md
```diff
@@ -97,13 +102,23 @@ scat = """
     """
 sco(scat)
 ```
+<!--
+Sorry, but I don't like this example, because the same could be achieved much more easily.
+- `cmap` is not used -> either add a comment or text what this *could* be good/used for or delete it.
+- Either state 3 colors in in the `ColorScheme` and omit the `values` parameter of `cgrad` OR
+  state 2 colors in the `ColorScheme` and use the `values` parameter of `cgrad`.
+  Using both is redundant.
```
lazarusA, line 110: in cgrad you always need to indicate how many colors you want. So, the goal here is to give the user several options to do a Colorbar and see what do you actually need. It seem to have worked as you already noticed. Good idea regarding bad, neutral and good I will take this into account later.

contents/data_vis_makie_latex.md
```diff
@@ -30,6 +30,11 @@ A more involved example will be one with some equation as `text` and increasing
 ```jl
 @sco JDS.multiple_lines()
 ```
+<!--
+The `lines` labels can be "simplified" to `L"%$i x"`. Then there also would be
```
lazarusA, line 34: Even though LaTeXStrings if loaded with Makie I want the package loading to remain explicit. That way if the user wants can google it and go directly to that package if needed.

contents/data_vis_makie_layouts.md
```diff
@@ -166,6 +172,11 @@ Then, the `inset` is easily done, as in:
 ```jl
 @sco JDS.figure_box_inset()
 ```
+<!--
+At least change `right` of `inset_ax2` or add a protrusion so the xaxis label
+isn't cropped. Alternatively enlarge thr `right` value on purpose to make the
```
lazarusA, line 177: that's exactly the point! in this scenario the inset plot extends more than needed it. It's kinda obvious, because is very visual.

lazarusA (approved):
Thank you for all your suggestions and taking the time to do them. I will merge this as it is and  I will come back later to your other comments.

contents/data_vis_makie_attributes.md
Mo-Gul: As I stated in the next sentence, I'd prefer italics. I think you have used that in other places as well. When I also remember you used bold for highlighting *statements* (that one should remember).

contents/data_vis_makie_colors.md
Mo-Gul: That doesn't seem to be right, because

```jl
using Colors, ColorSchemes
using CairoMakie
CairoMakie.activate!()

figure = (; resolution=(600, 400), font="CMU Serif")
axis = (; xlabel=L"x", ylabel=L"y", aspect=DataAspect())
mygrays = ColorScheme([Gray(i) for i in [0.0, 0.5, 1.0]])
fig, _, pltobj = heatmap(
    rand(-1:1, 20, 20);
    colormap=cgrad(mygrays, categorical=true, rev=true),
    axis=axis,
    figure=figure,
)
cbar = Colorbar(fig[1, 2], pltobj, label="Categories")
cbar.ticks = ([-0.66, 0, 0.66], ["bad", "neutral", "good"])
fig
```

works perfectly fine for me (in Julia v1.7). This is supported by the output of `?cgrad()` which gives

```jl
cgrad(colors, [values]; categorical = nothing, scale = nothing, rev = false, alpha = nothing)
```

and I guess the arguments in square brackets are optional.

contents/data_vis_makie_latex.md
Mo-Gul: That is perfectly fine for me. Nonetheless you should use the "simplified" version (everywhere) and optimally add a sentence why in LaTeXStrings variables (`$i`) need to be prefixed with `%`, i.e. `%$i`. (To distinguish it from the start of a mathematical expression.)

contents/data_vis_makie_layouts.md
Mo-Gul: I'd say that "more than needed" is quite subjective ...  
And just that we are on the same page: On my screen the missing part of the inset is less then half a zero of the xticklabel 10. This is the case for the PDF as well as the website. And _less than half a zero_ at least for me is not "kinda obvious, because is very visual". That is why I suggest to move the inset "far" into the margin that minimum the whole "10" isn't visible.

Mo-Gul:
And just to be sure that this was intended:  
Since you have _merged_ the PR also all my comments are now in the _main_ branch ...

PR #237, opened 2021-12-23
revision makie chapter

lazarusA:


storopoli (approved):

src/makie.jl
```diff
@@ -333,21 +333,21 @@ function mixed_mode_layout()
     CairoMakie.activate!() # hide
     seed!(123)
     longlabels = ["$(today() - Day(1))", "$(today())", "$(today() + Day(1))"]
-    fig = Figure(resolution=(600, 400), fontsize=12,
-        backgroundcolor=:grey90, font="CMU Serif")
-    ax1 = Axis(fig[1, 1])
-    ax2 = Axis(fig[1, 2], xticklabelrotation=pi / 2, alignmode=Mixed(bottom=0),
-        xticks=([1, 5, 10], longlabels))
+    fig = Figure(resolution = (600, 400), fontsize = 12,
```
rikhuijzer, line 336: Could we stick to no spacing around equal symbols in keyword arguments? I know that Makie has the convention to use the spacing, but in most places in the Julia community no spacing is used. See for example https://github.com/invenia/BlueStyle#keyword-arguments.

Otherwise, we need to add the spacing in the earlier chapters too to keep things consistent (and less confusing) for the reader

rikhuijzer: Apart from that nice work on the revision :+1:

PR #238, opened 2021-12-23
Use `@code_llvm` to explain performance

rikhuijzer:
This should make that section a lot clearer

### DIFF
--- a/contents/why_julia.md
+++ b/contents/why_julia.md
@@ -136,22 +136,76 @@ We don't know any other serious language that would want to be hard and slow, so
 **Julia is fast!
 Very fast!**
 It was designed for speed from the beginning.
-It accomplishes this by multiple dispatch.
+In the rest of this section, we go into details about why this is.
+If you don't have (much) programming experience yet, feel free to skip to the next section and maybe come later to this at a later moment.
+
+Julia accomplishes it's speed partially due to multiple dispatch.
 Basically, the idea is to generate very efficient LLVM[^LLVM] code.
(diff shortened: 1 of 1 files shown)


contents/why_julia.md
```diff
@@ -136,22 +136,76 @@ We don't know any other serious language that would want to be hard and slow, so
...
+top:
+    %1 = shl i64 %0, 1
+    %2 = add i64 %1, 3
+    ret i64 %2
+}
+```
+
+This is low-level LLVM code showing that the program only does the following:
+
+1. shift the input (3) one bit to the left, which has the same effect as multiplying by 2
```
storopoli, line 194:
```suggestion
1. shift the input (3) one bit to the left, which has the same effect as multiplying by 2; and
```

contents/why_julia.md
```diff
@@ -136,22 +136,76 @@ We don't know any other serious language that would want to be hard and slow, so
...
+1. add 3
```
storopoli, line 195:
```suggestion
1. add 3.
```

contents/why_julia.md
```diff
@@ -136,22 +136,76 @@ We don't know any other serious language that would want to be hard and slow, so
...
+
+and that's it!
+Julia has realized that calling `inner` can be removed, so that's not part of the calculation anymore!
+Now, imagine that this function is called a thousand or even a million times.
+These optimizations will reduce the running time significantly.
```
storopoli, line 200:
```suggestion
These optimizations would significantly reduce the running time.
```

storopoli (approved):
Nice addition.

PR #239, opened 2021-12-23
Add ISBN to CITATION.cff

rikhuijzer:
The only place that I could add the ISBN was in the preferred citation area (https://github.com/citation-file-format/citation-file-format/blob/main/examples/1.2.0/pass/key-complete/CITATION.cff), so that's why I added that.

I have no idea what the precise effects are on Zotero but I guess it is an improvement.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -24,7 +24,8 @@ keywords:
   - "Julia language"
   - "programming"
   - "software"
-  - "graphics"
+  - "data manipulation"
+  - "data visualization"
 
 license: "CC-BY-NC-SA-4.0"
 
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,27 +1,47 @@
-# YAML 1.2
----
+# Based on https://github.com/citation-file-format/citation-file-format/blob/main/examples/1.2.0/pass/key-complete/CITATION.cff.
+
+cff-version: "1.2.0"
+message: "If you use this software, please cite it as below."
+
 authors:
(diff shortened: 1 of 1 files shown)


CITATION.cff
```diff
@@ -1,27 +1,47 @@
...
-  -
-    type: url
+  - type: url
     value: "https://juliadatascience.io/"
+
+keywords:
+  - "Julia language"
+  - "programming"
+  - "software"
+  - "graphics"
```
storopoli, line 27:
```suggestion
  - "data manipulation"
  - "data visualization"
```

CITATION.cff
```diff
@@ -1,27 +1,47 @@
...
 license: "CC-BY-NC-SA-4.0"
-message: "If you use this software, please cite it using these metadata."
+
 repository-code: "https://github.com/JuliaDataScience/JuliaDataScience"
+
 title: "Julia Data Science"
+
 version: 1.0.0
+
+preferred-citation:
```
storopoli, line 37: Don't we have to put `authors` here also in the `preferred-citation`?

storopoli (approved):

CITATION.cff
rikhuijzer: I have no idea, I was hoping that systems would pull it from the layer up if the info is missing. If we merge, then I can check what Zotero makes from it

storopoli: Ok! Let's try that.

PR #251, opened 2022-01-27
Fixing buttons in JDS

storopoli:
The same as https://github.com/JuliaDataScience/JuliaDataScience-PT/pull/16 but here also.


rikhuijzer:
Not necessary because JDS doesn't set `online_url_prefix` in `config.toml`.

storopoli:
Thanks!

PR #252, opened 2022-01-28
Add link to Chinese version

rikhuijzer:
I've asked @findmyway and he said we can add the link


storopoli:
We also need to change the `index.md` at the `CN` version.

PR #261, opened 2022-05-07
Update Makie

rikhuijzer:
I've updated the Makie.jl versions at the Books.jl side (https://github.com/JuliaBooks/Books.jl/pull/278 and https://github.com/JuliaBooks/Books.jl/pull/279). This PR

- Closes #258
- Closes #259 
- Closes #260 

I'm just gonna merge if CI passes. I'll check the site for mistakes once it is online and fix things if necessary.


rikhuijzer:
Ah yes. Of course. Blocked by https://github.com/JuliaPlots/AlgebraOfGraphics.jl/pull/386.

rikhuijzer:
@lazarusA The inset plot can only take `:top`, `:center`, `:right` instead of numbers, so 
image

changed to

image

Is that okay?

lazarusA:
ohh... that's bad.... the idea is to be able to put the inset plot inside, any location. If this is the only thing that works, for now is ok. I will need to look for another solution later then.

Issue #263, opened 2022-05-09
The inset plot can only take `:top`, `:center`, `:right` instead of numbers


storopoli:
@lazarusA The inset plot can only take `:top`, `:center`, `:right` instead of numbers, so 
![image](https://user-images.githubusercontent.com/20724914/167393480-c8638e32-219a-417f-9148-42a0e69a3394.png)

changed to

![image](https://user-images.githubusercontent.com/20724914/167393495-b67f3375-0c6f-4d72-bffa-2206bced7d15.png)

Is that okay?

_Originally posted by @rikhuijzer in https://github.com/JuliaDataScience/JuliaDataScience/issues/261#issuecomment-1120933150_

PR #264, opened 2022-05-11
Revision makie chapter

lazarusA:
some clean up, plus this should also close https://github.com/JuliaDataScience/JuliaDataScience/issues/263 by means of using `aligmode`, the previous approach is no longer available, breaking change.


rikhuijzer (approved):
Awesome that you fixed this

I think it would be good to update the book on Amazon to avoid people reading old code. Are you okay with me updating the Amazon book once this PR is merged?

storopoli:
Hey folks, I only see code changes. Isn't it suppose to have also "prose" changes  in the `.md` files?

Issue #266, opened 2022-05-13
Discuss `Makie.jl` vs `Plots.jl`


storopoli:
> Many packages for Julia use plotting package Plots instead of Makie because of the recipe facility in Plots. Some discussion of how lack of recipes in current Makie is dealt with, would have been useful.

Source: [Amazon Review by B. Lie](https://www.amazon.com/Julia-Data-Science-Jose-Storopoli/dp/B09KMRKQ96)

PR #269, opened 2022-05-22
simple recipe for a DataFrame

lazarusA:
this adds a new section to the book, so I would say that is a breaking change. On the other hand I believe that is necessary to continue adding new content. Please take a look and suggest changes, specially for the DataFrames bit. 

This also addresses https://github.com/JuliaDataScience/JuliaDataScience/issues/266 very little, but some 😄 .

### DIFF
--- a/contents/data_vis_makie_recipe.md
+++ b/contents/data_vis_makie_recipe.md
@@ -27,7 +27,7 @@ s = """
 sc(s)
 ```
 
-note that here the macro `@recipe` will automatically create for us two new functions, `dfplot` and `dfplot!`, all lowercase from our type `DfPlot`.
+Note that the macro `@recipe` will automatically create two new functions for us, `dfplot` and `dfplot!`, all lowercase from our type `DfPlot`.
 The first one will create a complete new figure whereas the second one will plot into the current axis or an axis of your choosing.
 This allows us to plot `DataFrame`s which contains columns named, `x`, `y`, `z`.
 Now, let's take care of our plot definition.
--- a/contents/data_vis_makie_recipe.md
+++ b/contents/data_vis_makie_recipe.md
@@ -1,0 +1,96 @@
+## A Makie recipe for a DataFrame {#sec:recipe_df}
+
+Unlike other libraries that already support a wide set of input formats via recipes, i.e. `Plots.jl`, in `Makie.jl` most of the time we need to pass the raw data to our functions.
+However, if we want to define our own `recipe` in Makie we could also do it.
+As an example we will code a small **full recipe** for a `DataFrame`. Please refer to the [documentation](https://makie.juliaplots.org/stable/documentation/recipes/) for more details.
+
+A Makie `recipe` consist of two parts, a plot `type` name defined via `@recipe` and a custom `plot!`(`Makie.plot!`) which creates the actual plot via plotting functions already defined.
+
+```jl
+s = """
+    @recipe(DfPlot, df) do scene
+        Attributes(
+            x = :A,
+            y = :B,
+            c = :C,
+            color = :red,
+            colormap = :plasma,
+            markersize = 20,
+            marker = :rect
+        )
+    end
+    """
+sc(s)
+```
+
+with this we will have `dfplot` and `dfplot!` conveniently defined for us.
+Now, let's take care of our plot definition. We will do a simple scatter plot.
+
+```
+import Makie
+```
+
+```jl
+s = """
+    function Makie.plot!(p::DfPlot{<:Tuple{<:DataFrame}})
+        df = p[:df][]
+        x = getproperty(df, p[:x][])
+        y = getproperty(df, p[:y][])
+        c = getproperty(df, p[:c][])
+        scatter!(p, x, y; color = c, markersize = p[:markersize][],
+            colormap = p[:colormap][], marker = p[:marker][])
+        return p
+    end
+    """
+sc(s)
+```
+
+Note the extras `[]` at the end of each variable. Those are due to the fact that *recipes* in Makie are dynamic, meaning that our plots will update if our variables change. See [observables](https://makie.juliaplots.org/stable/documentation/nodes/) to know more.
+Now, we apply our new plotting function to the following `DataFrame`:
+
+```jl
+s = """
+    df_recipe = DataFrame(A=randn(10), B=randn(10), C=rand(10))
+    """
+sc(s)
+```
+
+```jl
+s = """
+    CairoMakie.activate!() # hide
+    dfplot(df_recipe)
+    label = "dfRecipe" # hide
+    caption = "DataFrames recipe." # hide
+    link_attributes = "width=60%" # hide
+    Options(current_figure(); filename=label, caption, label, link_attributes) # hide
+    """
+sco(s)
+```
+
+The named attributes in the recipe allows us to pass custom names to our new plotting function. Namely,
+
+```jl
+s = """
+    df_names = DataFrame(a1=rand(100), a2=rand(100), a3=rand(100))
+    """
+sc(s)
+```
+
+and
+
+```jl
+s = """
+    CairoMakie.activate!() # hide
+    dfplot(df_names; x = :a1, y = :a2, c = :a3, marker = 'o',
+        axis = (; aspect=1, xlabel = "a1", ylabel = "a2"),
+        figure = (; backgroundcolor = :grey90))
+    label = "dfRecipeArgs" # hide
+    caption = "DataFrames recipe with arguments." # hide
+    link_attributes = "width=60%" # hide
+    Options(current_figure(); filename=label, caption, label, link_attributes) # hide
+    """
+sco(s)
+```
+
+Note, that now we are calling by name each column as well as the marker type, allowing us to use this definition for different DataFrames.
+Additionally, all our previous options, i.e., `axis` or `figure` also work!
--- a/contents/data_vis_makie_recipe.md
+++ b/contents/data_vis_makie_recipe.md
@@ -4,7 +4,7 @@ Unlike other libraries that already support a wide set of input formats via reci
 However, we can also define our own  `recipe` in  `Makie.jl`.
 A `recipe` is your own custom plotting type command.
 This extension is done just in `Makie.jl`, which means that making a new set of plotting rules for your own types is light, namely, you don't need the complete plotting machinery available to define them. This is specially useful if you want to include your own plotting commands in one of your own packages.
-However, in order for them to work you will still need to use one of the backends, i.e., GLMakie.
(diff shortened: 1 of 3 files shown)


contents/data_vis_makie_recipe.md
```diff
@@ -0,0 +1,96 @@
+## A Makie recipe for a DataFrame {#sec:recipe_df}
+
+Unlike other libraries that already support a wide set of input formats via recipes, i.e. `Plots.jl`, in `Makie.jl` most of the time we need to pass the raw data to our functions.
+However, if we want to define our own `recipe` in Makie we could also do it.
```
rikhuijzer, line 4:
```suggestion
However, we can also define our own  `recipe` in  `Makie.jl`.
```

contents/data_vis_makie_recipe.md
```diff
@@ -0,0 +1,96 @@
...
+As an example we will code a small **full recipe** for a `DataFrame`. Please refer to the [documentation](https://makie.juliaplots.org/stable/documentation/recipes/) for more details.
```
rikhuijzer, line 5: Can you elaborate a bit more on what a recipe is? To be honest, I'm also not sure whether I fully understand it. As far as I understand, recipes are meant for lightweight interface packages so that you can define plotting in your package without having to load the full plotting library? Also, it allows defining a basic framework?

contents/data_vis_makie_recipe.md
```diff
@@ -0,0 +1,96 @@
...
+
+A Makie `recipe` consist of two parts, a plot `type` name defined via `@recipe` and a custom `plot!`(`Makie.plot!`) which creates the actual plot via plotting functions already defined.
```
rikhuijzer, line 7:
```suggestion
A Makie `recipe` consist of two parts, a plot `type` name defined via `@recipe` and a custom `plot!(::Makie.plot)` which creates the actual plot via plotting functions already defined.
```

contents/data_vis_makie_recipe.md
```diff
@@ -0,0 +1,96 @@
...
+            colormap = :plasma,
+            markersize = 20,
+            marker = :rect
+        )
+    end
+    """
+sc(s)
+```
+
+with this we will have `dfplot` and `dfplot!` conveniently defined for us.
```
rikhuijzer, line 26:
```suggestion
with this we will have `dfplot` and `dfplot!` conveniently defined for us.
With this, we can conveniently plot `DataFrame`s which contains columns named, `x`, `y`, `z`.
```

contents/data_vis_makie_recipe.md
```diff
@@ -0,0 +1,96 @@
...
+        c = getproperty(df, p[:c][])
+        scatter!(p, x, y; color = c, markersize = p[:markersize][],
+            colormap = p[:colormap][], marker = p[:marker][])
+        return p
+    end
+    """
+sc(s)
+```
+
+Note the extras `[]` at the end of each variable. Those are due to the fact that *recipes* in Makie are dynamic, meaning that our plots will update if our variables change. See [observables](https://makie.juliaplots.org/stable/documentation/nodes/) to know more.
```
rikhuijzer, line 48: Ah, is it because the DfPlot is immutable and they use pointers to (de)reference stuff? Sounds a bit dangerous but okay

rikhuijzer (approved):
Some small comments. Adding new stuff is great :+1:

contents/data_vis_makie_recipe.md
```diff
@@ -0,0 +1,96 @@
+## A Makie recipe for a DataFrame {#sec:recipe_df}
+
+Unlike other libraries that already support a wide set of input formats via recipes, i.e. `Plots.jl`, in `Makie.jl` most of the time we need to pass the raw data to our functions.
```
storopoli, line 3:
```suggestion
Unlike other libraries that already support a wide set of input formats via recipes, i.e. `Plots.jl`, in `Makie.jl` most of the time we need to pass the raw data to functions.
```

contents/data_vis_makie_recipe.md
storopoli: Yeah I agree with @rikhuijzer, recipes need at least a simple definition.

contents/data_vis_makie_recipe.md
storopoli: "With this" is duplicated as starting point for two times in a row on your suggestion @rikhuijzer.

contents/data_vis_makie_recipe.md
storopoli, line 48:
```suggestion
Note the extras `[]` at the end of each variable.
Those are due to the fact that *recipes* in Makie are dynamic, meaning that our plots will update if our variables change.
See [observables](https://makie.juliaplots.org/stable/documentation/nodes/) to know more.
```

Just a better semantic break here.

contents/data_vis_makie_recipe.md
```diff
@@ -0,0 +1,96 @@
...
+    dfplot(df_recipe)
+    label = "dfRecipe" # hide
+    caption = "DataFrames recipe." # hide
+    link_attributes = "width=60%" # hide
+    Options(current_figure(); filename=label, caption, label, link_attributes) # hide
+    """
+sco(s)
+```
+
+The named attributes in the recipe allows us to pass custom names to our new plotting function. Namely,
```
storopoli, line 70:
```suggestion
The named attributes in the recipe allows us to pass custom names to our new plotting function. Namely:
```

contents/data_vis_makie_recipe.md
```diff
@@ -0,0 +1,96 @@
...
+
+```jl
+s = """
+    df_names = DataFrame(a1=rand(100), a2=rand(100), a3=rand(100))
+    """
+sc(s)
+```
+
+and
```
storopoli, line 79:
```suggestion
and:
```

contents/data_vis_makie_recipe.md
```diff
@@ -0,0 +1,96 @@
...
+            markersize = 20,
+            marker = :rect
+        )
+    end
+    """
+sc(s)
+```
+
+with this we will have `dfplot` and `dfplot!` conveniently defined for us.
+Now, let's take care of our plot definition. We will do a simple scatter plot.
```
storopoli, line 27:
```suggestion
Now, let's take care of our plot definition.
We will do a simple scatter plot:
```

storopoli (commented):
Some suggestions.

contents/data_vis_makie_recipe.md
rikhuijzer: Fixed

contents/data_vis_makie_recipe.md
lazarusA: I also agree on this 😄 ... I will come back shortly.

contents/data_vis_makie_recipe.md
lazarusA: is the whole logic of observables in Makie 😄

contents/data_vis_makie_recipe.md
lazarusA: An extra paragraph is up.

rikhuijzer (approved):

contents/data_vis_makie_recipe.md
```diff
@@ -0,0 +1,104 @@
+## A Makie recipe for a DataFrame {#sec:recipe_df}
+
+Unlike other libraries that already support a wide set of input formats via recipes, i.e. `Plots.jl`, in `Makie.jl` most of the time we need to pass the raw data to functions.
+However, we can also define our own  `recipe` in  `Makie.jl`.
+A `recipe` is your own custom plotting type command.
+This extension is done just in `Makie.jl`, which means that making a new set of plotting rules for your own types is light, namely, you don't need the complete plotting machinery available to define them. This is specially useful if you want to include your own plotting commands in one of your own packages.
+However, in order for them to work you will still need to use one of the backends, i.e., GLMakie.
```
storopoli, line 7:
```suggestion
However, in order for them to work you will still need to use one of the backends, i.e., GLMakie or CairoMakie.
```

contents/data_vis_makie_recipe.md
```diff
@@ -0,0 +1,104 @@
+## A Makie recipe for a DataFrame {#sec:recipe_df}
+
+Unlike other libraries that already support a wide set of input formats via recipes, i.e. `Plots.jl`, in `Makie.jl` most of the time we need to pass the raw data to functions.
+However, we can also define our own  `recipe` in  `Makie.jl`.
+A `recipe` is your own custom plotting type command.
```
storopoli, line 5: I still don't think this is clear.
For instance, a new user (the profile of our reader) would be confused.

Does a recipe create a `plot` and `plot!` custom function for my custom types?
Like call i simply call:

```julia
plot(my_custom_type)
```

and it will plot?

It is not clear.
Down below you create a recipe for a `DfPlot`:

```julia
@recipe(DfPlot, df) do scene
...
```

But somehow you have to call it with `dfplot` (all lowercase).
Our reader would be asking themselves "WTF, how does that work? Do I have to use `plot[!]`, or `dfplot`? And how did it got converted to a lowercase plotting function?"

storopoli (changes requested):
I still don't think this is clear.
For instance, a new user (the profile of our reader) would be confused.

Does a recipe create a `plot` and `plot!` custom function for my custom types?
Like call i simply call:

```julia
plot(my_custom_type)
```

and it will plot?

It is not clear.
Down below you create a recipe for a `DfPlot`:

```julia
@recipe(DfPlot, df) do scene
...
```

But somehow you have to call it with `dfplot` (all lowercase).
Our reader would be asking themselves "WTF, how does that work? Do I have to use `plot[!]`, or `dfplot`? And how did it got converted to a lowercase plotting function?"

contents/data_vis_makie_recipe.md
lazarusA: agree. Now it saids that the macro `@recipe` is responsible for creating two new lowercase functions that you are suppose to use later on.

contents/data_vis_makie_recipe.md
```diff
@@ -27,7 +27,8 @@ s = """
 sc(s)
 ```
 
-with this we will have `dfplot` and `dfplot!` conveniently defined for us.
+note that here the macro `@recipe` will automatically create for us two new functions, `dfplot` and `dfplot!`, all lowercase from our type `DfPlot`.
```
rikhuijzer, line 30:
```suggestion
Note that the macro `@recipe` will automatically create two new functions for us, `dfplot` and `dfplot!`, all lowercase from our type `DfPlot`.
```

rikhuijzer (approved):

contents/data_vis_makie_recipe.md
lazarusA: great! thanks a lot @rikhuijzer

storopoli (approved):

PR #272, opened 2022-08-12
Switch to Books v2

rikhuijzer:
I'm having a bit of trouble running this locally (NixOS nor Docker manage to run all blocks :sweat_smile:), but if CI passes then all should be good. Unless there are objections, I'm planning to send an updated PDF to Amazon because syntax highlighting is disabled for output blocks now which looks better.

### `Books.jl` v2 release notes

Version 2 closes about 6 long-standing issues. Most noteworthy:

- Syntax highlighting is disabled for output blocks in both HTML and PDF. This makes the books and websites look much cleaner.
- The output is printed via `ProgressMeter.jl` instead of a dump to the console. This avoids a lot of scrolling when using `gen`.
- All code is now always evaluated in `Main`. There used to be two ways to do it, but having one is much simpler and therefore better.
- Yggdrasil is now used for Tectonic so Tectonic will automatically update with Yggdrasil and supports ARM via their build system (pandoc-crossref doesn't yet unfortunately)
- docx is removed because, well, it's docx
- `install_dependencies` is avoided by passing JuliaMono directly into the TeX file. This avoids the need for a globally installed JuliaMono

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,7 +17,7 @@ jobs:
 
       - uses: julia-actions/setup-julia@v1
         with:
-          version: "1.7"
+          version: "1"
 
       - name: Copy over the Portuguese output
         shell: bash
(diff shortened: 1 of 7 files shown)


rikhuijzer:
More info on the GLMakie error in https://github.com/JuliaPlots/Makie.jl/issues/2214. This error was unrelated to this PR, but I think that `size(xyz, 2)` should have been `size(xyz, 1)`. Sounds correct @lazarusA?

.github/workflows/CI.yml
```diff
@@ -17,7 +17,7 @@ jobs:
 
       - uses: julia-actions/setup-julia@v1
         with:
-          version: "1.7"
+          version: "1"
```
storopoli, line 20: Nice that is a good change with 1.8 right around the corner.

storopoli (approved):

PR #276, opened 2022-10-20
Use `eachindex` instead of `1:length`

jariji:
`1:length` depends on the endpoints of the indexing, so is unsafe for OffsetArrays. `eachindex` is more generic and clear about the intent of the operation.

### DIFF
--- a/contents/julia_basics.md
+++ b/contents/julia_basics.md
@@ -679,7 +679,7 @@ For example, we can create a function that adds 1 to each element in a vector `V
 ```jl
 s = """
     function add_one!(V)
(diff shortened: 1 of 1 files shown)


rikhuijzer:
I don’t like it super much because this section is mostly aimed at beginners and `1:length(…)` is more intuitive. Having said that, you’re right. It’s best if people would use `eachindex`, so I’ll merge

thank you for opening a PR!

storopoli:
@jariji thanks for the PR, but we want to keep the book simple for beginners. Most of them have no idea what `OffsetArrays.jl` are.

PR #280, opened 2022-11-14
Update Makie to 0.18

rikhuijzer:
Closes:

- #273 
- #274 
- #275 

I've built the book from scratch again with these updated `Makie.jl` versions. The were only some warnings and zero errors. I also manually compared the PDF of the online version with the locally built PDF for the data visualization chapter and the output is exactly the same.

I expect that this PR can be merged if CI passes. @lazarusA can you confirm?


lazarusA:
I'm surprise that that syntax in Makie still works. I will do some changes later as a Christmas update. For now let's go ahead and merge.

PR #281, opened 2022-11-14
Add support button

### LINKED ISSUE #277: add an add to the physical book on Amazon
I was thinking that maybe having an add (beside the links in the first page) to the book in the website will be nice to have.

rikhuijzer:
- Closes #277 

## Desktop Preview

desktop preview

## Mobile Preview

image

Let's merge when both reviewers are happy


lazarusA:
is too hidden at the bottom. Could we might have it at the top? and use a [fa-fa coffee] ? cup in green? or other colour, but not without it.

lazarusA:
I see. My only issue is that is not visible enough. Maybe putting a box around it [in green, red, orange or], like in the one you mentioned in the issue?, I could take a look at later on how to include it in your span element.

storopoli (approved):
Once @lazarusA is fine with the placement and details we can merge.

metadata.yml
```diff
@@ -5,7 +5,16 @@ author:
   - Jose Storopoli
   - Rik Huijzer
   - Lazaro Alonso
-html-license: '<a href="http://creativecommons.org/licenses/by-nc-sa/4.0/">CC BY-NC-SA 4.0</a>'
+html-license: >-
+  <span style="display: flex; justify-content: center;">
```
lazarusA, line 9:
```suggestion
 <span style="width: 200px; height: 35px; display: flex; justify-content: center; border: 1.5px solid rgb(67, 72, 54);border-radius: 10px;">
```
we could go for red, green or completely black, now is kinda dark and probably is already ok. Any colour will do, feel free to merge @rikhuijzer, with the suggestion or without it.

lazarusA (approved):
please, take a look at the suggestion.

PR #287, opened 2022-12-03
Update Makie to 0.19

rikhuijzer:


rikhuijzer:
I'm gonna merge since this, according to Julius:

> First, there’s a breaking change: All occurrences of textsize are now fontsize, so search and replace that word and you will be good to go. 

As I read it (https://discourse.julialang.org/t/ann-makie-jl-0-19/91184), only the change of this word is breaking and the rest is only new features. So, it should be safe to just move to the new version.

PR #298, opened 2023-01-05
cheatSheet_cairo.jl improvements

knuesel:
Some suggested changes to the CairoMakie cheatsheet, some for consistency and some to make it easier to understand what the function actually does (I think that is the main use of this figure: once the purpose of a plotting function is clear, the user can always check the documentation for the different ways to call the function). The biggest change is for `linesegments`: it now uses the `linesegments(x, y)` signature with the same data as the previous plots, to help understanding what's going on.

Full list of changes:

* change range of first plots to have even number of points (for linesegments)
* change linesegments to use same data as previous plots
* uniformize parameter names in plot titles
* use variable heights in crossbar
* fix title of violin plot
* more explicit title for mesh

### DIFF
--- a/src/cheatSheet_cairo.jl
+++ b/src/cheatSheet_cairo.jl
@@ -8,7 +8,7 @@ function cheatsheet_cairomakie()
     packages = ["Makie", "CairoMakie"]
     seed!(123)
     CairoMakie.activate!()
-    x = 0:0.5:2π
+    x = range(0, 2π, length=12)
     fig = with_theme(theme_light()) do
         fig = Figure(resolution=(1000, 1600))
         axs = [Axis(fig[i, j], aspect=1) for i = 1:7 for j = 1:5]
@@ -25,10 +25,8 @@ function cheatsheet_cairomakie()
         stem!(axs[4], x, sin.(x); color=x)
         axs[4].title = "stem(x, y)"
 
-        linesegments!(axs[5],
-            Point2f.(vec([[x, rand()] for i = 1:2, x in rand(10)]));
-            color=1:20, colormap=:gnuplot)
-        axs[5].title = "linesegments(positions)"
+        linesegments!(axs[5], x, sin.(x); color=1:length(x)/2, colormap=:gnuplot)
+        axs[5].title = "linesegments(x, y)"
 
         series!(axs[6], rand(10, 5); color=resample_cmap(:plasma, 10))
         axs[6].title = "series(curves)"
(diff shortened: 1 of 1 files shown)


src/cheatSheet_cairo.jl
```diff
@@ -25,10 +25,8 @@ function cheatsheet_cairomakie()
         stem!(axs[4], x, sin.(x); color=x)
         axs[4].title = "stem(x, y)"
 
-        linesegments!(axs[5],
-            Point2f.(vec([[x, rand()] for i = 1:2, x in rand(10)]));
-            color=1:20, colormap=:gnuplot)
-        axs[5].title = "linesegments(positions)"
```
lazarusA, line 31: thanks for your suggestions. However, for linesegments, this approach was on purpose, since is not in the official docs and is the use case that I find myself the most useful. See here for example: https://beautiful.makie.org/dev/examples/generated/2d/datavis/astronauts/ and the original version actually 😄 https://github.com/lazarusA/100daysOfMakie#4-linesegments

src/cheatSheet_cairo.jl
```diff
@@ -89,14 +87,15 @@ function cheatsheet_cairomakie()
...
+
         xbox = rand(1:3, 1000)
         boxplot!(axs[21], xbox, randn(1000), color=xbox,
             mediancolor=:black, colormap=[:grey, :dodgerblue, :yellow])
         axs[21].title = "boxplot(x, y)"
 
         violin!(axs[22], xbox, randn(1000), color=:black,
             show_median=true, mediancolor=:white)
-        axs[22].title = "boxplot(x, y)"
+        axs[22].title = "violin(x, y)"
```
lazarusA, line 98: nice catch!

lazarusA (changes requested):
I kinda like my original example for linesegments, if your have an improve suggestion I could take a look again, but your current suggestion avoids my points about using a list of points.

src/cheatSheet_cairo.jl
knuesel: I think the cheatsheet is most useful for showing "what the plotting functions do".  The piecewise sinewave works great for this, because the reader can see how the same points from the previous plots are connected by segments in this plot. And using the simplest method (`linesegments(x, y)`) is also easier to understand.

It's true that this example doesn't show the most interesting way to use `linesegments`. But I don't think the cheatsheet should focus on "how to do things with this function" because anyway there are many ways to use a Makie function, and the best way depends on the use case.

What about this: we show a simple case in the cheatsheet and we add some less simple examples to the official documentation of `linesegments`. Would that be OK for you? If yes I'm happy to make a Makie PR to add some examples there.

Or if you prefer, I can replace `linesegments(x, sin.(x))` in my example with `linesegments(Point2f.(x, sin.(x))`. This uses the same method (vector of `Point2f`) as in your example.

knuesel: I've pushed another commit using `linesegments(Point2f.(x, sin.(x))`.

knuesel:
I'v added a commit to use a list of points, let me know if you'd like other changes.

PR #296, opened 2022-12-25
create a figure again.

lazarusA:
@rikhuijzer why I keep getting: 
```
pandoc-crossref: Duplicate label: fig:fig_axis_plot
```
this is in the new .md file `*create_figure.md`

PR #299, opened 2023-01-26
Put the cheat sheet on a separate page

rikhuijzer:
This will make the cheat sheets available at <https://juliadatascience.io/makie_cheat_sheet> which makes it easier to find via Google and also easier to link to.

What do you think, @lazarusA?


lazarusA:
yes. Good for me. I have done a lot of work in the draft, https://github.com/JuliaDataScience/JuliaDataScience/pull/296. Unfortunately, I haven't sprint into the finish line.

PR #312, opened 2023-06-19
Edit for clarity

simonp0420:
The original wording can be confusing to a naïve reader.  For example, the phrase "Columns are usually of the same data type" sounds like it is saying that columns have a data type and this type is usually the same for all columns.  I think it is clearer to state that all the elements in a given column share the same data type.


rikhuijzer:
Thanks, Peter! I agree. It's clearer now

PR #319, opened 2023-11-25
ci: add dependabot

storopoli:
Also updates some old actions that were giving warnings on CI. Changes the runner machines to all be ubuntu-latest which is the latest LTS ubuntu.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: daily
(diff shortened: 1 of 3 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: daily
```
rikhuijzer, line 6:
```suggestion
      interval: weekly
```

rikhuijzer (approved):
Thanks for looking into this!

.github/dependabot.yml
storopoli: Yes, daily is overkill

Issue #321, opened 2023-11-25
CI: Upload PDF as the CI Artifact


storopoli:
We have no idea if the PDF generated on CI is valid, this would be a good idea.

PR #322, opened 2023-11-25
Provide PDF artifact in PR

rikhuijzer:
Fixes https://github.com/JuliaDataScience/JuliaDataScience/issues/321.

I'm still debugging so feel free to unsubscribe.


storopoli:
Yak shaving. I think that the feature is correctly implemented and it works.
Only that the PDF is not building!

storopoli (approved):
it works!

PR #323, opened 2023-11-26
fix(makie): warnings from 0.20 migration

storopoli:
The only one I didn't fixed was `Warning: argument help isn't currently implemented correctly` since this is hardcoded in Makie (https://github.com/MakieOrg/Makie.jl/blob/822d11cf1e9751084a276f79b616b03afdc43e2d/src/documentation/documentation.jl#L63)...


rikhuijzer (approved):
Thanks for fixing this, José!

PR #324, opened 2023-11-26
fix(docker): update Dockefile to julia:latest

storopoli:
This fix might not need a retouch in the future.


rikhuijzer (approved):
Ah, nice! Latest refers to the latest stable release according to Docker Hub: https://hub.docker.com/_/julia.

PR #326, opened 2023-11-28
ci: simplify formatting with julia-actions/julia-format

storopoli:
This simplifies the format CI and delegates everything to [`julia-actions/julia-format`](https://github.com/julia-actions/julia-format). With this workflow, it will automatically post code suggestions to pull requests, based on the formatting rules defined by JuliaFormatter.jl and JuliaFormatter.toml file.


.github/workflows/Format.yml
```diff
@@ -1,36 +1,12 @@
 name: Format
 
 on:
-  push:
-    branches:
-      - main
```
rikhuijzer, line 6: Why not? What if someone pushes straight to `main` and then inside a PR someone else is surprised that the formatter fails

.github/workflows/Format.yml
```diff
@@ -1,36 +1,12 @@
...
   pull_request:
   workflow_dispatch:
 
 jobs:
-  docs:
-    name: CheckFormatting
+  code-style:
```
rikhuijzer, line 8:
```suggestion
  format:
```
maybe?

rikhuijzer (commented):
Have you verified that the new approach detects the same formatting problems as the old approach? Does juliaformatter manually go through Markdown files nowadays and does it check for trailing whitespace?

storopoli:
I have changed my mind. I think that what we have works....

PR #330, opened 2023-12-18
Reduce scope

rikhuijzer:
This is not a good day I'm having. So much noise. This is the last one.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -11,7 +11,7 @@ jobs:
   BuildAndDeploy:
     runs-on: ubuntu-latest
     permissions:
-      issues: write
+      pull-requests: write
     env:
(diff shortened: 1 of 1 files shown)


storopoli:
What's the difference?

rikhuijzer:
> What's the difference?

More restrictive permissions. It doesn’t matter since this project open source so I don’t see much that could go wrong, but in general it’s good to keep permissions as tight as possible. Especially if someone might copy our code.

The reason it suddenly started crashing is unclear to me. GitHub decided to move from the old security system to the new. Maybe related to the new interface? I don’t know. It’s fixed now

storopoli:
> Maybe related to the new interface? I don’t know. It’s fixed now

Thanks!

Issue #345, opened 2024-08-18
Cairomakie cheat sheet has huge file size and appears empty


rikhuijzer:
@lazarusA do you maybe know why https://juliadatascience.io/makie_cheat_sheets shows no image for the CairoMakie one? Also the front cover is empty. In the PDF, the image does work:

![image](https://github.com/user-attachments/assets/98db8943-0061-48d1-bb9c-3da797ad98be)

Could it be that one is in PNG format and the web in SVG?

lazarusA:
I will take a look.

PR #346, opened 2024-08-18
La/makie issues

lazarusA:
this should close https://github.com/JuliaDataScience/JuliaDataScience/issues/345

### DIFF
--- a/src/cheatSheet_cairo.jl
+++ b/src/cheatSheet_cairo.jl
@@ -7,7 +7,7 @@ end
 function cheatsheet_cairomakie()
     packages = ["Makie", "CairoMakie"]
     seed!(123)
-    CairoMakie.activate!()
+    CairoMakie.activate!(type="png")
     x = range(0, 2π; length=12)
     fig = with_theme(theme_light()) do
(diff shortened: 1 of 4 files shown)


rikhuijzer:
Thanks!

rikhuijzer:
I couldn't test locally. You fixed the front page indeed 🥳 🥳 

The cheat sheet is still broken, but I'll fix that. The PNG works, so I will just add a workaround that changes the SVG to PNG extension in the HTML output

rikhuijzer:
Everything seems to work again. Thanks a lot for the help @lazarusA!

lazarusA:
Screenshot 2024-08-19 at 15 33 55

this one also looks off. I will take a look later.

rikhuijzer:
> Screenshot 2024-08-19 at 15 33 55
> 
> this one also looks off. I will take a look later.

Yep same for me. Probably same problem :( It works in the PDF

lazarusA:
yes, we should change also output to `png`.

rikhuijzer:
> yes, we should change also output to `png`.

And maybe a Makie bug report? What do you think?

PR #348, opened 2024-10-10
Set Julia version to 1.11

rikhuijzer:


rikhuijzer:
Looking at the PDF, there are still some regressions since the switch to Typst (https://github.com/JuliaBooks/Books.jl/pull/321), but apart from that everything seems to work.

storopoli:
Why not pin to 1.10 since that is the new LTS?

PR #349, opened 2024-10-11
Set Julia version to 1.10

rikhuijzer:
In response to

> Why not pin to 1.10 since that is the new LTS?

https://github.com/JuliaDataScience/JuliaDataScience/pull/348#issuecomment-2405773369


storopoli:
Thanks!

PR #365, opened 2025-09-16
Make a BonitoBook based version

SimonDanisch:
This is a work in progress.
Right now it contains a BonitoBook style and plugin, which allows to turn many markdowns into a complete book as a website, where each page is a BonitoBook.
I removed all Books.jl based syntax, but that still needs a rigorous review, to see what's missing and incorrect now.
I also still need to run all the cells when building, which I expect to fail for a few cells right now.
I want to switch to Bonito's build infrastructure to support a preview of the finished site, and I want to make a few other goodies, like the language switches as part of the overall design - we then should also have a download button, to run the whole book interactively on the users PC.


SimonDanisch:
I put the code to run / static export the whole book into `make.jl`.

SimonDanisch:
Thank you :) 
Possibly too early, if you ask me, unless you already want to see in what direction this is going!
But I will definitely do a few more iteration until this is running smoothly - I want to make sure that everything runs at least and produces a PR preview with the completely build site.

storopoli:
> want to see in what direction this is going!

mainly this yeah.

The review will probably be more looking at the old book vs the new book than Julia code anyways 😂

storopoli:
Is there a way to have nested headers in the left banner table of contents?

